PR #25, opened 2021-10-09
allow fuseki config per service (query, update and graphStore) and ad…

### LINKED ISSUE #26: Extend fuseki configuration per service and add Authentication
At the moment fuseki configuration just takes the URL of the endpoint. Add option to extend this with 1 URL per service (query, update and graphStore).
Also add the option to use a password protected fuseki service.

domivds:
…d auth support

### DIFF
--- a/cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/FusekiConfig.java
+++ b/cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/FusekiConfig.java
@@ -1,0 +1,65 @@
+package zone.cogni.libs.sparqlservice.impl;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class FusekiConfig extends Config {
+
+  private String updateUrl;
+  private String queryUrl;
+  private String graphStoreUrl;
+  private String overwriteTurtleMimeType;
+
+  public static FusekiConfig from(Config config) {
+    FusekiConfig fusekiConfig = new FusekiConfig();
+    fusekiConfig.setUrl(config.getUrl());
+    fusekiConfig.setUser(config.getUser());
+    fusekiConfig.setPassword(config.getPassword());
+    return fusekiConfig;
+  }
+
+  public String getUpdateUrl() {
+    return getServiceUrl(updateUrl, "/update");
+  }
+
+  public FusekiConfig setUpdateUrl(String updateUrl) {
+    this.updateUrl = updateUrl;
+    return this;
+  }
+
+  public String getQueryUrl() {
+    return getServiceUrl(queryUrl, "/query");
+  }
+
+  public FusekiConfig setQueryUrl(String queryUrl) {
+    this.queryUrl = queryUrl;
+    return this;
+  }
+
+  public String getGraphStoreUrl() {
+    return getServiceUrl(graphStoreUrl, "/data");
+  }
+
+  public FusekiConfig setGraphStoreUrl(String graphStoreUrl) {
+    this.graphStoreUrl = graphStoreUrl;
+    return this;
+  }
+
+  public String getTurtleMimeType() {
+    return StringUtils.defaultIfBlank(overwriteTurtleMimeType, "text/turtle");
+  }
+
+  public String getOverwriteTurtleMimeType() {
+    return overwriteTurtleMimeType;
+  }
+
+  public FusekiConfig setOverwriteTurtleMimeType(String overwriteTurtleMimeType) {
+    this.overwriteTurtleMimeType = overwriteTurtleMimeType;
+    return this;
+  }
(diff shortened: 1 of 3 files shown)


cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/FusekiSparqlService.java
```diff
@@ -17,42 +27,65 @@
 import java.io.IOException;
 import java.io.StringWriter;
 import java.nio.charset.StandardCharsets;
+import java.util.Base64;
 import java.util.Collections;
 import java.util.function.Function;
 
 public class FusekiSparqlService implements SparqlService {
```
natan-cox-cognizone, line 34: why not use `ConfigurationProperties` so configuration is part of the component?

we could get rid of FusekiConfig this way.

natan-cox-cognizone: I'm not even sure where and how `SparqlServiceProvider` get executed.

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/FusekiConfig.java
```diff
@@ -0,0 +1,65 @@
+package zone.cogni.libs.sparqlservice.impl;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class FusekiConfig extends Config {
+
+  private String updateUrl;
+  private String queryUrl;
+  private String graphStoreUrl;
```
natan-cox-cognizone, line 9: are those 3 urls so different and unpredictable? can they not be based on the same "root" url?

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/FusekiSparqlService.java
```diff
@@ -76,35 +107,55 @@ public void executeUpdateQuery(String updateQuery) {
 
   @Override
   public void upload(Model model, String graphUri) {
```
alexey-lukashov, line 111: is this upload replacing graph or updating with new values? one method we made in virtuoso became very handy with replace graph on demand parameter `public void upload(Model model, String graphUri, boolean replace)`

alexey-lukashov (approved):

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/FusekiSparqlService.java
domivds: true
I propose we add that add as a future improvement. At same time when we do some clean up and deprecate one of the 2 sparql service versions.
Note, the advantage of the Provider is that we pass the type in the config. So the bean creation uses the provider and doesn't use a specific SparqlService class.

In mopo we have this as usage:
```  @Bean
  @MopoDataSource(MopoDataSource.Source.rdfstoreCuriaInput)
  public SparqlService sparqlServiceCuriaInput() {
    return sparqlServiceProvider().createSparqlService(MopoDataSource.Source.rdfstoreCuriaInput);
  }
```
  
  MopoDataSource.Source enum has 2 values (we use 2 different fuseki stores).

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/FusekiConfig.java
domivds: In fuseki you can configure your different services with any URL you want. And this happened in mopo...

By default you can still just pass the main endpoint URL and the config will calculate the default values.

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/FusekiSparqlService.java
domivds: good point
pushed an update for this

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/FusekiSparqlService.java
natan-cox-cognizone: Cleanup of 2 sparql services and clean up of for services are configured can be handled separately.

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/FusekiConfig.java
natan-cox-cognizone: weird. ok then 👍

natan-cox-cognizone (approved):

PR #29, opened 2021-10-17
feature: added TemplateService interface

### LINKED ISSUE #30: Create interface for processing templates
SpelService might be replaced in some cases with different implementations

alexey-lukashov:
TemplateService interface to support different implementations of template service

### DIFF
--- a/cube/src/main/java/zone/cogni/asquare/cube/spel/SpelService.java
+++ b/cube/src/main/java/zone/cogni/asquare/cube/spel/SpelService.java
@@ -12,27 +12,10 @@ import zone.cogni.core.spring.ResourceHelper;
 import java.util.HashMap;
 import java.util.Map;
 
-public class SpelService {
+public class SpelService implements TemplateService {
 
   private final SpelExpressionParser spelExpressionParser = new SpelExpressionParser();
   private final Map<String, Expression> expressionCache = new HashMap<>();
-
-  public NamedTemplate processTemplate(NamedTemplate namedTemplate, Object root) {
-    String result = processTemplate(namedTemplate.getTemplate(), root);
-    return getNewNamedTemplate(namedTemplate, root, result);
-  }
-
-  private NamedTemplate getNewNamedTemplate(NamedTemplate namedTemplate, Object root, String result) {
-    NamedTemplate copy = namedTemplate.copy();
-    copy.setRoot(root);
-    copy.setResult(result);
-    return copy;
-  }
-
-  public String processTemplate(Resource templateResource, Object root) {
-    String template = ResourceHelper.toString(templateResource);
-    return processTemplate(template, root);
-  }
 
   public String processTemplate(String template, Object root) {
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@
 /build
 /*/build
 /*/out
+/application-profile/collapse-single-ap-output.ap.json
```
natan-cox-cognizone: hmm. yes. I think this one is produced by one of tests. should we not check if it can be avoided.

cube/src/main/java/zone/cogni/asquare/cube/spel/SpelService.java
```diff
@@ -12,23 +12,11 @@
...
-  private NamedTemplate getNewNamedTemplate(NamedTemplate namedTemplate, Object root, String result) {
-    NamedTemplate copy = namedTemplate.copy();
-    copy.setRoot(root);
-    copy.setResult(result);
-    return copy;
-  }
-
   public String processTemplate(Resource templateResource, Object root) {
     String template = ResourceHelper.toString(templateResource);
     return processTemplate(template, root);
```
natan-cox-cognizone: looks like this could be also a default method on the interface (unless there is a special reason for not doing it)

.gitignore
natan-cox-cognizone: so in `CollapseApplicationProfileTest`

let's remove 
```
   //    JsonNode jsonNode = applicationProfileConfig.getSerializer().apply(collapse);
    JsonNode jsonNode = new ApplicationProfileSimpleJson().apply(collapse);

    print(jsonNode);
```

and also 
```
private void print(JsonNode jsonNode) {
    try {
      ObjectMapper mapper = new ObjectMapper();
      ObjectWriter writer = mapper.writer(new DefaultPrettyPrinter());

      File file = new FileSystemResource("collapse-single-ap-output.ap.json").getFile();
      writer.writeValue(file, jsonNode);
    }
    catch (IOException e) {
    }
  }
```

cube/src/main/java/zone/cogni/asquare/cube/spel/TemplateService.java
```diff
@@ -0,0 +1,23 @@
...
+    String processTemplate(String template, Object root);
+
+    String processTemplate(Resource templateResource, Object root);
+
+    default NamedTemplate processTemplate(NamedTemplate namedTemplate, Object root) {
+        String result = processTemplate(namedTemplate.getTemplate(), root);
+        return getNewNamedTemplate(namedTemplate, root, result);
+    }
+
+    default NamedTemplate getNewNamedTemplate(NamedTemplate namedTemplate, Object root, String result) {
```
natan-cox-cognizone: this was a private method but java 8 does not allow this. maybe `getNamedTemplateCopy` is more clear?

natan-cox-cognizone (changes requested):

natan-cox-cognizone (approved):

PR #40, opened 2021-11-26
ElasticsearchIndexingService implementation, no unit tests (yet) but …

### LINKED ISSUE #39: Configurable indexing service
When indexing into Elasticsearch facets is already picked up in a generic way. Also, multiple components are already in place to handle the rest of the indexing, but there is no generic way of doing it.

The proposal is to make a service which handles every step of the flow

- create and delete of indexes
- calculation of URIs to be indexed
- loading of RDF to be indexed
- conversion to JSON
- calculation of facets

natan-cox-cognizone:
…tested on an existing application

### DIFF
--- a/cube/src/main/java/zone/cogni/asquare/cube/index/CollectionMetadata.java
+++ b/cube/src/main/java/zone/cogni/asquare/cube/index/CollectionMetadata.java
@@ -1,0 +1,49 @@
+package zone.cogni.asquare.cube.index;
+
+import org.springframework.core.io.Resource;
+
+import java.util.List;
+
+class CollectionMetadata {
+
+  private String name;
+  private List<Resource> selectQueryResources;
+  private List<Resource> constructQueryResources;
+  private List<Resource> facetQueryResources;
+
+  public String getName() {
+    return name;
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
+
+  public List<Resource> getSelectQueryResources() {
+    return selectQueryResources;
+  }
+
+  public void setSelectQueryResources(List<Resource> selectQueryResources) {
+    this.selectQueryResources = selectQueryResources;
+  }
+
+  public List<Resource> getConstructQueryResources() {
+    return constructQueryResources;
+  }
+
+  public void setConstructQueryResources(List<Resource> constructQueryResources) {
+    this.constructQueryResources = constructQueryResources;
+  }
+
+  public List<Resource> getFacetQueryResources() {
+    return facetQueryResources;
+  }
+
+  public void setFacetQueryResources(List<Resource> facetQueryResources) {
+    this.facetQueryResources = facetQueryResources;
+  }
+
+  public boolean isValid() {
+    return !selectQueryResources.isEmpty() && !constructQueryResources.isEmpty();
+  }
+}
--- a/cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
+++ b/cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
@@ -214,7 +214,7 @@ public class ElasticsearchIndexingService {
   }
 
   private void indexByCollection(IndexMetadata indexMetadata, List<String> collections) {
-    log.info("(indexByCollection) index '" + indexMetadata.getName() + "' and collections: " + String.join(", ", collections));
+    log.info("(indexByCollection) index '{}' and collections: {}", indexMetadata.getName(), String.join(", ", collections));
     List<Callable<String>> callables = getCallables(indexMetadata, collections).collect(Collectors.toList());
 
     log.info("(indexByCollection) {} uris found", callables.size());
--- a/cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
+++ b/cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
@@ -1,0 +1,604 @@
+package zone.cogni.asquare.cube.index;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.vavr.control.Try;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.ModelFactory;
+import org.apache.jena.rdf.model.RDFNode;
+import org.elasticsearch.index.IndexNotFoundException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.context.ApplicationContext;
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.core.io.FileSystemResource;
+import org.springframework.core.io.Resource;
+import zone.cogni.asquare.cube.convertor.ModelToJsonConversion;
+import zone.cogni.asquare.cube.monitoredpool.MonitoredPool;
+import zone.cogni.asquare.cube.pagination.PaginatedQuery;
+import zone.cogni.asquare.cube.sparql2json.SparqlSelectToJson;
+import zone.cogni.asquare.cube.spel.SpelService;
+import zone.cogni.asquare.service.elasticsearch.v7.Elasticsearch7Store;
+import zone.cogni.asquare.triplestore.RdfStoreService;
+
+import javax.annotation.Nonnull;
+import javax.annotation.PostConstruct;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Here is an expected folder structure for indexing:
+ *
+ * <pre>
+ *   /rootFolder                <- root folder
+ *     /name-1                  <- name of index
+ *       elastic-settings.json
+ *       /collection-1-1        <- name of collection
+ *         select-*.sparql      <- select uris for collection
+ *         construct-*.sparql   <- select triples for uri
+ *         /facets              <- folder for facet queries
+ *           select-1.sparql    <- facet query
+ *           select-2.sparql
+ *       /collection-1-2
+ *         select-*.sparql
+ *         ...
+ *     /name-2
+ *       elastic-settings.json
+ *       /collection-2-1
+ *         select-uris.sparql
+ *         ...
+ * </pre>
+ */
+public class ElasticsearchIndexingService {
+
+  private static final Logger log = LoggerFactory.getLogger(ElasticsearchIndexingService.class);
+
+  private static final String indent = "        ";
+
+  private final ApplicationContext resourcePatternResolver;
+  private final String configurationClasspath;
+
+  private final SpelService spelService;
+  private final PaginatedQuery paginatedQuery;
+  private final MonitoredPool indexMonitoredPool;
+
+  private final RdfStoreService rdfStore;
+  private final Elasticsearch7Store elasticStore;
+  private final ModelToJsonConversion modelToJsonConversion;
+
+  private List<IndexMetadata> indexesMetadata;
+
+  /**
+   * Common parameters used in queries, like sparql endpoint urls etc...
+   */
+  private final Map<String, String> queryTemplateParameters;
+
+  public ElasticsearchIndexingService(String configurationClasspath,
+                                      Map<String, String> queryTemplateParameters,
+                                      SpelService spelService,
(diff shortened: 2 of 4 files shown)


cube/src/main/java/zone/cogni/asquare/cube/index/CollectionMetadata.java
```diff
@@ -0,0 +1,49 @@
...
+  public List<Resource> getFacetQueryResources() {
+    return facetQueryResources;
+  }
+
+  public void setFacetQueryResources(List<Resource> facetQueryResources) {
+    this.facetQueryResources = facetQueryResources;
+  }
+
+  public boolean isValid() {
+    return !selectQueryResources.isEmpty() && !constructQueryResources.isEmpty();
```
domivds, line 47: also add null checks and check if name is filled in?
facetQueryResources is non-mandatory, so i would or initialize with empty List or check for non-null here

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
```diff
@@ -0,0 +1,604 @@
...
+
+  private List<Resource> getCollectionSelectQueries(String index, String collection) {
+    return Arrays.asList(getResources("/" + index + "/" + collection + "/select-*.*"));
+  }
+
+  private List<Resource> getCollectionConstructQueries(String index, String collection) {
+    return Arrays.asList(getResources("/" + index + "/" + collection + "/construct-*.*"));
+  }
+
+  @PostConstruct
```
domivds: Proposal: put @PostConstruct on top of the class after the constructor

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
```diff
@@ -0,0 +1,604 @@
...
+ *         select-*.sparql
+ *         ...
+ *     /name-2
+ *       elastic-settings.json
+ *       /collection-2-1
+ *         select-uris.sparql
+ *         ...
+ * </pre>
+ */
+public class ElasticsearchIndexingService {
```
domivds, line 54: Global class proposal: split up in "real indexing" and "init/find config files" classes.

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
```diff
@@ -0,0 +1,604 @@
...
+
+    return configurationClasspath.endsWith("/") ? configurationClasspath.substring(0, configurationClasspath.length() - 1)
+                                                : configurationClasspath;
+  }
+
+  private List<IndexMetadata> calculateIndexesMetadata() {
+    Resource[] resources = getResources("/**/*");
+    return getLocalPaths(resources).stream()
+                                   .map(this::stripSlashAtFront)
+                                   .filter(path -> path.contains("/")) // <- make sure it's a folder
```
domivds: possible issue in windows, can be fixed with something like this
if(File.separatorChar != '/') path = path.replace(File.separatorChar, '/');

not sure what happens if a system with non-/ separator has a / in the filename
anyway, anyone doing that should be punished

note: can be fixed in the getLocalPath method (in the FileSystemResource case) so we only use /

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
```diff
@@ -0,0 +1,604 @@
...
+
+    indexesMetadata.forEach(this::validateIndex);
+  }
+
+  private void validateIndex(IndexMetadata indexMetadata) {
+    // valid or not
+    if (indexMetadata.isValid())
+      log.info("  '{}' index is valid", indexMetadata.getName());
+    else
+      log.error("  '{}' index NOT is valid", indexMetadata.getName());
```
domivds: just log error? no hard fail?
we could do all validations and if 1 fails then stop at the end so we have the full report

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
```diff
@@ -0,0 +1,604 @@
...
+    log.info("{}   collection count: {}", indent, collectionNames.size());
+
+    if (collectionNames.isEmpty()) {
+      log.error("{}   no valid collections configured", indent);
+    }
+
+    if (!collectionNames.isEmpty())
+      log.info("{}   valid collections:   {}", indent, String.join(", ", collectionNames));
+    if (!getInvalidCollectionNames(indexMetadata).isEmpty())
+      log.info("{}   invalid collections: {}", indent, String.join(", ", getInvalidCollectionNames(indexMetadata)));
```
domivds: error log?

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
```diff
@@ -0,0 +1,604 @@
...
+  public List<String> getIndexNames() {
+    return indexesMetadata.stream()
+                          .filter(IndexMetadata::isValid)
+                          .map(IndexMetadata::getName)
+                          .collect(Collectors.toList());
+  }
+
+  private List<String> getInvalidIndexNames() {
+    return indexesMetadata.stream()
+                          .filter(indexMetadata -> !indexMetadata.isValid())
```
domivds: Fun one (might be added in later java version)
add method somewhere:
```  
public static <T> Predicate<T> not(Predicate<T> t) {
    return t.negate();
}
  ```
  
then do this:
`.filter(not(IndexMetadata::isValid))`

File: cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java [hunk 8]
```diff
@@ -0,0 +1,604 @@
...
+  }
+
+  /**
+   * Fills a single index.
+   * Can do a clear index in <code>clear</code> parameter is set to <code>true</code>.
+   *
+   * @param index to be filled
+   * @param clear clears index before filling it again
+   */
+  private void indexByName(@Nonnull String index, boolean clear) {
```
domivds: should be public i guess

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
```diff
@@ -0,0 +1,604 @@
...
+                      .flatMap(collectionMetadata -> getCallables(indexMetadata, collectionMetadata));
+  }
+
+  /**
+   * @param indexMetadata      of index
+   * @param collectionMetadata of object uris being loaded in index
+   * @return stream of <code>Callable</code>s for a single collection in an index
+   */
+  private Stream<Callable<String>> getCallables(IndexMetadata indexMetadata, CollectionMetadata collectionMetadata) {
+    log.info("(getCallables) for index '" + indexMetadata.getName() + "' and collection '" + collectionMetadata.getName() + "'");
```
domivds: log template

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
```diff
@@ -0,0 +1,604 @@
...
+   * @return Callable instance based on uri
+   */
+  private Callable<String> getCallable(IndexMethod indexMethod, List<Resource> collectionConstructQueries, String uri) {
+    Supplier<Model> modelSupplier = getModelSupplier(collectionConstructQueries, uri);
+    return indexMethod.indexOneCallable(modelSupplier, uri);
+  }
+
+  private Supplier<Model> getModelSupplier(List<Resource> collectionConstructQueries, String uri) {
+    return () -> {
+      Map<String, String> createModelMap = getCreateModelMap(uri);
```
domivds: rename createModelMap  to templateParameterMap or so?

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
```diff
@@ -0,0 +1,604 @@
...
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Here is an expected folder structure for indexing:
+ *
+ * <pre>
+ *   /rootFolder                <- root folder
+ *     /name-1                  <- name of index
```
domivds, line 36: future update: have a mapping per environment, so we can have same folder structure but a different index name per environment (or per installation)

cube/src/main/java/zone/cogni/asquare/cube/index/CollectionMetadata.java
natan-cox-cognizone: I invoke setters during initialization of data. can lookup for resources return null and not an empty list (or array)? @domivds

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
natan-cox-cognizone: yes. will be ok after moving code.

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
natan-cox-cognizone: if we would do this, I think it would be better look at spring boot options here and not a DIY solution. for now I would consider this a new feature and create an issue.

natan-cox-cognizone: on the other hand: bad configuration also just crashes. will add it here.

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
natan-cox-cognizone: error is already logged before: this is just informational in the context of an index.

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
natan-cox-cognizone: weird you have to do it yourself. in jdk11 there is a method btw.

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
natan-cox-cognizone: not sure. for APIs it's not needed. they can easily use the list solution. but if we think it could be useful I can do it.

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
natan-cox-cognizone: oops

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
natan-cox-cognizone: see issue #41

cube/src/main/java/zone/cogni/asquare/cube/index/CollectionMetadata.java
domivds: ah, class is package private
then all fine 👍

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
domivds: for maintenance this can be useful (ofc if there is an api calling that), so i propose to make it public

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
```diff
@@ -0,0 +1,368 @@
...
+                           .get();
+  }
+
+  public void indexByCollection(String index, List<String> collections) {
+    IndexMetadata indexMetadata = getIndexMetadata(index);
+    indexByCollection(indexMetadata, collections);
+  }
+
+  private void indexByCollection(IndexMetadata indexMetadata, List<String> collections) {
+    log.info("(indexByCollection) index '" + indexMetadata.getName() + "' and collections: " + String.join(", ", collections));
```
domivds:
```suggestion
    log.info("(indexByCollection) index '{}' and collections: {}", indexMetadata.getName(), String.join(", ", collections));
```

domivds (approved):

cube/src/main/java/zone/cogni/asquare/cube/index/ElasticsearchIndexingService.java
natan-cox-cognizone: yes. I changed it. strange that it does not show up here.

PR #46, opened 2021-12-20
implementation of hot swap for indexing

### LINKED ISSUE #45: Elasticsearch indexing should support hot swapping

natan-cox-cognizone:


asquareroot/src/main/java/zone/cogni/asquareroot/elastic/EmbeddedElasticsearch7Store.java
```diff
@@ -47,7 +47,7 @@
   private static final Logger log = LoggerFactory.getLogger(EmbeddedElasticsearch7Store.class);
   private final ObjectMapper objectMapper = new ObjectMapper();
   private Node node;
-  private Client сlient;
+  private Client client;
```
alexey-lukashov, line 50: merci:)

asquareroot/src/main/java/zone/cogni/asquareroot/elastic/EmbeddedElasticsearch7Store.java
```diff
@@ -93,6 +93,13 @@ public boolean isClosed() {
     return node.isClosed();
   }
 
+
+  @Override
+  public String getUrl() {
+    // not sure if this is correct
+    return "http://127.0.0.1:9200/";
```
alexey-lukashov, line 100: should be fine

cube/src/main/java/zone/cogni/asquare/cube/index/InternalIndexingServiceUtils.java
```diff
@@ -0,0 +1,146 @@
...
+  @Nonnull
+  static IndexMethod getIndexMethodForUri(@Nonnull IndexingServiceContext context,
+                                          @Nonnull String indexToFill,
+                                          @Nonnull List<Resource> facetQueryResources,
+                                          @Nonnull String uri) {
+    Resource[] queryResources = facetQueryResources.toArray(new Resource[0]);
+    SparqlSelectToJson sparqlSelectToJson = new SparqlSelectToJson(queryResources,
+                                                                   context.getSpelService(),
+                                                                   getTemplateParameterMap(context, uri));
+    return new IndexMethod(context.getPaginatedQuery(),
```
alexey-lukashov, line 89: @natan-cox-cognizone what do you think about ading here a Function<Model, String> for externalized computable facets? for some complicated cases when we would need java to produce facet json with java?

alexey-lukashov (approved):

cube/src/main/java/zone/cogni/asquare/cube/index/InternalIndexingServiceUtils.java
natan-cox-cognizone: @alexey-lukashov Yes. For content facets, we are going to need something more pluggable. 

Also, the current setup is probably not too good. API looks a bit weird and performance is suffering because we always need to "read" Resource inputstream.

This is definitely an area which requires improvement.

alexey-lukashov:
I would try to use this solution.
This approach requires to create indexing queries (select/construct/facet) manually... but feels like we will have more understanding on what is selected and why it is indexed at the end. 
Construct query has potential flexibility, in 95% simple cases data could be computed in sparql. I guess this would resolve code complexity of pluggable facet services of our existing indexing implementation.

And Im not going to blame existing approach which is based on A2 and application profiles, it is performing automated generation of select/construct queries and json transformation according to AP.
Probably automated approach might be also useful for some express implementation of indexing, especially when we want to index simple but very nested structures. And I might be wrong but... as far as I know, we dont really use multiple level nested indexing, instead we are trying to avoid it for performance reasons. That's why personally for me this manual approach seems more attractive.

cube/src/main/java/zone/cogni/asquare/cube/index/InternalIndexingServiceUtils.java
alexey-lukashov: also might be useful to add timestamp for every indexed resource, and implement garbage collector based on elastic delete query, its risky but super fast deleting obsolete stuff

natan-cox-cognizone: @alexey-lukashov created a new issue https://github.com/cognizone/asquare/issues/50

natan-cox-cognizone:
> 

Note: construct selects triples to be added in the JSON document (which can be ?s ?p ?o for a single graph) but conversion still uses the same principle of converting triples to JSON and it uses exactly the same format.

The main reason for adding a construct query for the selection of triples is the use case where we use triples and not graphs. But it can be useful when using graphs too.

cube/src/main/java/zone/cogni/asquare/cube/index/swap/IndexSwapService.java
```diff
@@ -0,0 +1,134 @@
...
+import static zone.cogni.asquare.cube.index.swap.JsonBuilder.object;
+
+public class IndexSwapService {
+
+  private static final Logger log = LoggerFactory.getLogger(IndexSwapService.class);
+
+  private final ElasticStore elasticStore;
+  private final ElasticsearchMetadataService elasticsearchMetadataService;
+
+  private final RestTemplate restTemplate = new RestTemplate(clientHttpRequestFactory(3000, 10000));
```
domivds, line 27: can we come up with some way of having this in a default config that we can overwrite per project

cube/src/main/java/zone/cogni/asquare/cube/index/swap/JsonBuilder.java
```diff
@@ -0,0 +1,197 @@
...
+
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import javax.annotation.Nonnull;
+
+class JsonBuilder {
```
domivds, line 11: seems reusable, can it be put in some json lib package?

service/src/main/java/zone/cogni/asquare/service/elasticsearch/info/ElasticsearchMetadataService.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.http.client.ClientHttpRequestFactory;
+import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
+import org.springframework.web.client.RestTemplate;
+import zone.cogni.asquare.service.elasticsearch.ElasticStore;
+
+public class ElasticsearchMetadataService {
+
+  private static final Logger log = LoggerFactory.getLogger(ElasticsearchMetadataService.class);
+
+  private final RestTemplate restTemplate = new RestTemplate(clientHttpRequestFactory(3000, 3000));
```
domivds, line 15: same config remark as above

domivds (approved):

cube/src/main/java/zone/cogni/asquare/cube/index/swap/IndexSwapService.java
natan-cox-cognizone: true. I am definitely not the HTTP in java expert. @domivds what is best?

cube/src/main/java/zone/cogni/asquare/cube/index/swap/JsonBuilder.java
natan-cox-cognizone: once we clean it, it definitely an option

PR #51, opened 2021-12-27
add support for stardog

domivds:


cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/HttpHelper.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.http.client.CredentialsProvider;
+import org.apache.http.client.fluent.Response;
+import org.apache.http.client.protocol.HttpClientContext;
+import org.apache.http.impl.client.BasicAuthCache;
+import org.apache.http.impl.client.BasicCredentialsProvider;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+
+public class HttpHelper {
```
natan-cox-cognizone: public? or if we move it package private?

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/HttpHelper.java
```diff
@@ -0,0 +1,50 @@
...
+
+  public static HttpClientContext createAuthenticationHttpContext(String user, String password) {
```
natan-cox-cognizone, line 20: name of method does not really mention it is doing basic credential. not sure if there are "more secure" ways of creating a `HttpClientContext`

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/HttpHelper.java
```diff
@@ -0,0 +1,50 @@
...
+    context.setAuthCache(new BasicAuthCache()); //so after first call it will know it has to send the authentication
+    return context;
+  }
+
+  public static void checkAndDiscardResponse(Response response) throws IOException {
+    try {
+      HttpResponse httpResponse = response.returnResponse();
+      StatusLine statusLine = httpResponse.getStatusLine();
+      if (statusLine.getStatusCode() / 100 == 2) return;
+      throw new RuntimeException("Upload didn't answer 2xx code " + statusLine + ": " + getBody(httpResponse));
```
natan-cox-cognizone, line 45: I guess different backends return different error messages... do we need to add a translation in SparqlService? now it is basically, take response and dump it (which probably is good enough for exception messages)

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/StardogSparqlService.java
```diff
@@ -0,0 +1,158 @@
...
+              .setHeader(authHeader)
+              .bodyFile(file, ContentType.create("text/turtle", StandardCharsets.UTF_8))
+              .execute();
+      checkAndDiscardResponse(response);
+    }
+    catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+
+  }
```
natan-cox-cognizone:
```suggestion
    }
  }
```

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/StardogSparqlService.java
```diff
@@ -0,0 +1,158 @@
...
+  @Override
+  public boolean executeAskQuery(String askQuery) {
+    //pass default graph, method without default graph doesn't use the HttpContext !!!
+    try (QueryExecution queryExecution = QueryExecutionFactory.sparqlService(endpointUrl + "/query", askQuery, null, null, authenticationHttpContext)) {
+      //jena adds empty defaultGraph param to URL because defaultGraph is null but is a "value", stardog doesn't like that
+      ((QueryEngineHTTP) queryExecution).setDefaultGraphURIs(Collections.emptyList());
+      return queryExecution.execAsk();
+    }
+
+  }
```
natan-cox-cognizone:
```suggestion
    }
  }
```

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/StardogSparqlService.java
```diff
@@ -0,0 +1,158 @@
...
+  @Override
+  public void replaceGraph(String graphUri, Model model) {
+    upload(model, graphUri, true);
+  }
+
+  private void upload(Model model, String graphUri, boolean replace) {
+    String graphStoreUrl = endpointUrl + "?graph=" + graphUri;
+    StringWriter writer = new StringWriter();
+    try {
+      model.write(writer, "ttl");
```
natan-cox-cognizone: I would split of getting the string -> separate method

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/StardogSparqlService.java
```diff
@@ -0,0 +1,158 @@
...
+      Response response = (replace ? Request.Put(graphStoreUrl) : Request.Post(graphStoreUrl))  //Put replaces the graph, Post adds data
```
natan-cox-cognizone: wow :)
separate statement maybe

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/StardogSparqlService.java
```diff
@@ -0,0 +1,158 @@
...
+
+  }
+
+  @Override
+  public boolean executeAskQuery(String askQuery) {
+    //pass default graph, method without default graph doesn't use the HttpContext !!!
+    try (QueryExecution queryExecution = QueryExecutionFactory.sparqlService(endpointUrl + "/query", askQuery, null, null, authenticationHttpContext)) {
+      //jena adds empty defaultGraph param to URL because defaultGraph is null but is a "value", stardog doesn't like that
+      ((QueryEngineHTTP) queryExecution).setDefaultGraphURIs(Collections.emptyList());
+      return queryExecution.execAsk();
```
natan-cox-cognizone, line 82: why no `checkAndDiscardResponse`?

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/StardogSparqlService.java
```diff
@@ -0,0 +1,158 @@
...
+    try {
+      Response response = Request.Delete(graphStoreUrl)
+              .setHeader(authHeader)
+              .execute();
+      checkAndDiscardResponse(response);
+    }
+    catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
```
natan-cox-cognizone: a code block thingy that does something like this maybe? not sure how it works with checked exception though

```
try {
  Response response = supplier.get();
  checkAndDiscardResponse(response);
}
catch (IOException e) {
}  
```

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/HttpHelper.java
domivds: It creates an HttpClientContext that can authenticate with given username/password.
The authentication "system" will be handled by the httpClient lib. For example server can ask to authenticate via digest or basic auth. The lib will then kindof use the context to find the username/password.

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/HttpHelper.java
domivds: We can do this in the context of an in depth redesign of the sparql services. 
Here it's like we do for all the other SparqlService: 2xx code => success, any other code: fail and log why

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/StardogSparqlService.java
domivds: should be handled by the AutoCloseable

natan-cox-cognizone (approved):

PR #53, opened 2021-12-27
making timeouts for resttemplate for elasticsearch metadata services configurable

### LINKED ISSUE #52: clientHttpRequestFactory method is not really flexible
both in `IndexSwapService` and `ElasticsearchMetadataService`

natan-cox-cognizone:


service/src/main/java/zone/cogni/asquare/service/elasticsearch/info/ElasticsearchMetadata.java
```diff
@@ -122,4 +122,26 @@ public int hashCode() {
     }
   }
 
+  public static class Configuration {
+
+    private int connectTimeout;
+    private int readTimeout;
```
domivds: add a default config here
- user of the lib can work with default values if there is nothing specific in the users project
- if we add new config values, it won't break 
- we can still use @ConfigurationProperties to initialise only the values we want to overwrite

service/src/main/java/zone/cogni/asquare/service/elasticsearch/info/ElasticsearchMetadata.java
```diff
@@ -122,4 +122,26 @@ public int hashCode() {
...
+
+    public int getConnectTimeout() {
+      return connectTimeout;
+    }
+
+    public void setConnectTimeout(int connectTimeout) {
```
domivds, line 139: maybe set a bit a comment here (main info i want here is the unit, seconds/milliseconds?)

service/src/main/java/zone/cogni/asquare/service/elasticsearch/info/ElasticsearchMetadata.java
domivds:
```suggestion
    private int connectTimeout = 3000;
    private int readTimeout = 10000;
```

service/src/main/java/zone/cogni/asquare/service/elasticsearch/info/ElasticsearchMetadata.java
domivds, line 139:
```suggestion
    /**
    * Determines the timeout in milliseconds until a connection is established.
    * A timeout value of zero is interpreted as an infinite timeout. A negative value is interpreted as undefined (system default if applicable).
    * default = 3000
    **/
    public void setConnectTimeout(int connectTimeout) {
```

service/src/main/java/zone/cogni/asquare/service/elasticsearch/info/ElasticsearchMetadata.java
```diff
@@ -122,4 +122,26 @@ public int hashCode() {
...
+      this.connectTimeout = connectTimeout;
+    }
+
+    public int getReadTimeout() {
+      return readTimeout;
+    }
+
+    public void setReadTimeout(int readTimeout) {
```
domivds, line 152:
```suggestion
    /**
    * Defines the socket timeout (SO_TIMEOUT) in milliseconds, which is the timeout for waiting for data or, put differently, a maximum period inactivity between two consecutive data packets).
    * A timeout value of zero is interpreted as an infinite timeout. A negative value is interpreted as undefined (system default if applicable).
    * default = 1000
    **/
    public void setReadTimeout(int readTimeout) {
```

PR #58, opened 2022-01-19
feat: add more constructors for IndexMethod

marianna-tselepi:


cube/src/main/java/zone/cogni/asquare/cube/index/IndexMethod.java
```diff
@@ -67,6 +67,29 @@ public Params getParams() {
...
+
+  public IndexMethod(ModelToJsonConversion modelToJsonConversion,
+                     String indexName,
+                     Elasticsearch7Store elasticStore,
+                     SparqlSelectToJson sparqlSelectToJson) {
+    this(null, null, modelToJsonConversion, indexName, elasticStore, sparqlSelectToJson);
+  }
+
+  @Deprecated
+  public IndexMethod(PaginatedQuery paginatedQuery,
```
natan-cox-cognizone, line 84: @marianna-tselepi since we are deprecating use of RdfStoreService here, can we also deprecate all methods which have graphUri as a parameter. Those are related.

natan-cox-cognizone (changes requested):

natan-cox-cognizone (approved):

PR #60, opened 2022-02-04
Improve TDB handling

gabor-harsanyi:
- LocalTdbRdfStoreService
  - it was using model locking instead of transactions, it's replaced
  - its locking mechanism was implemented around Model and Graph, now it's Dataset based
  - it's internal closed state was not thread safe, allowing multiple closure and causing issue
  - added parameter validation
  - made getModel method Deprecated (because this can allow to move Model out of locking and out of transaction scope)
  - don't allow operations in closed state
  - made it poolable
  - add timeout on queries (to make sure they don't stuck)
  - changed read/write operations so now they all share the same locking/transaction context (Dataset)
    - Jena has one global transaction manager but several transactional view through transactional objects, like Dataset, Model, ...
-  Jena can have two way of working, either one shared global transaction or one single transactional view per thread, but it does not like to share transactional view among threads (that was the same with the model locks)
  - introduced object pooling so we don't spam the memory
  - the object pool is key based and each key can have a set of instances from the pool, it means one TDB can have several transactional view objects (TDB is the key)
  - added request context aware logic
    - to make sure only one transactional view per thread is present
    - object is returned back to the pool at the end of request
- Jena shares Connection to one TDB among all threads, all lock views and all transactional views
  - if any thread is interrupted (for example OOM) then Jena puts the connection into corrupted state and all other views will be invalidated
  - current pooling validates objects on several actions and also async periodically and removes invalid views
- Added test cases
- Added memory friendly result set handler


cogni-sem/src/main/java/zone/cogni/sem/jena/template/TooManyResultsException.java
```diff
@@ -0,0 +1,76 @@
...
+   * @param message            the detail message.
+   * @param cause              the cause.  (A {@code null} value is permitted,
+   *                           and indicates that the cause is nonexistent or unknown.)
+   * @param enableSuppression  whether or not suppression is enabled
+   *                           or disabled
+   * @param writableStackTrace whether or not the stack trace should
+   *                           be writable
+   *
+   */
+  public TooManyResultsException(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace) {
```
natan-cox-cognizone, line 73: @gabor-harsanyi so much comments... is this copied from somewhere?

gabor-harsanyi: They are generated constructors from RuntimeException

triplestore-jena-memory/src/test/resources/init_ctms_tdb/CTMS-SK-NSC-20221901.ttl
```diff
@@ -0,0 +1,3359 @@
+<http://data.europa.eu/esco/ctms/mapping/3b5f555b-21e9-4a5e-b8ea-41d7e7ce7750>
```
natan-cox-cognizone: not sure if we should put customer data in opensource?

cogni-sem/src/main/java/zone/cogni/sem/jena/template/TooManyResultsException.java
natan-cox-cognizone: Ah, wow. I did not even know this one :)

triplestore-jena-memory/src/test/resources/init_ctms_tdb/CTMS-SK-NSC-20221901.ttl
gabor-harsanyi: uh, yes, it's definitely a mistake, I'll replace, apology

gabor-harsanyi: buuut, as a second thought, this will be published along with esco data

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/pool/RdfStoreServiceProvider.java
```diff
@@ -0,0 +1,99 @@
...
+import zone.cogni.asquare.triplestore.pool.key.RdfStoreServicePoolKey;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+
+/**
+ * This should be one instance per thread and stores all TDBs connections that a given thread used.
+ * One connection per TDB btu one thread can use more TDBs.
```
natan-cox-cognizone:
```suggestion
 * One connection per TDB but one thread can use more TDBs.
```

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/pool/key/ConceptUriBasedPoolKey.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.io.File;
+import java.nio.file.Path;
+import java.util.Objects;
+
+public class ConceptUriBasedPoolKey implements RdfStoreServicePoolKey<String> {
+
+  private final Path dirPath;
+
+  public ConceptUriBasedPoolKey(final Path basePath, final String conceptUri) {
```
natan-cox-cognizone: slightly weird IMO. first of all not sure if we should limit ourselves to conceptUri.

second LocalTdbPoolKey seems remarkably similar. Except maybe the getKey function itself. And the fact that this class does not store the related uri.

gabor-harsanyi: The idea was to share the URI to "dir name" functionality with a base implementation and then each "real" implementation can inherit later from here. (because each real key needs to contain all the information what is needed to create a pooled rdf store service)
What would you suggest as a change?

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/pool/LocalTdbRdfStoreServicePool.java
```diff
@@ -0,0 +1,146 @@
...
+
+import java.io.Closeable;
+import java.time.Duration;
+
+public final class LocalTdbRdfStoreServicePool implements Closeable {
+
+  private static final Logger log = LoggerFactory.getLogger(LocalTdbRdfStoreServicePool.class);
+
+  private final GenericKeyedObjectPool<LocalTdbPoolKey, LocalTdbRdfStoreService> genericPool;
+  private final KeyedObjectPool<LocalTdbPoolKey, LocalTdbRdfStoreService> pool;
```
natan-cox-cognizone: I do not get the difference between the 2 pools

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/pool/key/ConceptUriBasedPoolKey.java
natan-cox-cognizone: Like I mentioned. I am not 100% sure what we are trying to achieve.  The parameter conceptUri is a bit confusing. How is it a conceptUri?

triplestore-jena-memory/src/test/resources/init_ctms_tdb/CTMS-SK-NSC-20221901.ttl
natan-cox-cognizone: I understand. But maybe we should keep a2 a bit customer neutral?

natan-cox-cognizone:
As a generic comment: I understand using a TDB has some possible issues compared to using a more simple or straightforward solution. What you did makes a lot of sense when an application uses a TDB. But it looks more like it should be a separate module instead of an implementation of the RdfStoreService interface.

You cannot swap in TDB just like another triplestore, imo.

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/pool/key/ConceptUriBasedPoolKey.java
gabor-harsanyi: what would you suggest to change it? merge the two classes?

gabor-harsanyi:
> As a generic comment: I understand using a TDB has some possible issues compared to using a more simple or straightforward solution. What you did makes a lot of sense when an application uses a TDB. But it looks more like it should be a separate module instead of an implementation of the RdfStoreService interface.
> 
> You cannot swap in TDB just like another triplestore, imo.

may I ask what would you suggest? new sub-module? same sub-module but new package? different repo?

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/pool/LocalTdbRdfStoreServicePool.java
gabor-harsanyi: genericPool is the not decorated pool and it is needed to be stored because it has some useful methods that we don't have on the interface, the `pool` is the final result, the decorated pool

triplestore-jena-memory/src/test/resources/init_ctms_tdb/CTMS-SK-NSC-20221901.ttl
gabor-harsanyi: I have pushed new test data

natan-cox-cognizone:
> may I ask what would you suggest? new sub-module? same sub-module but new package? different repo?

Not a different repo, almost all or "lib" code is here. I would start with a separate package. If we feel it can be useful it could be a submodule.

gabor-harsanyi:
> > may I ask what would you suggest? new sub-module? same sub-module but new package? different repo?
> 
> Not a different repo, almost all or "lib" code is here. I would start with a separate package. If we feel it can be useful it could be a submodule.

isn't it already a different package? I might misunderstand you

gabor-harsanyi:
> > may I ask what would you suggest? new sub-module? same sub-module but new package? different repo?
> 
> Not a different repo, almost all or "lib" code is here. I would start with a separate package. If we feel it can be useful it could be a submodule.

https://github.com/cognizone/asquare/pull/61

triplestore-tdb-pool/src/main/java/zone/cogni/asquare/triplestore/pool/factory/LocalTdbRdfStoreServiceFactory.java
```diff
@@ -0,0 +1,35 @@
...
+   *
+   * @param key the key used when constructing the object
+   *
+   * @return an instance that can be served by the pool
+   *
+   * @throws Exception if there is a problem creating a new instance,
+   *                   this will be propagated to the code requesting an object.
+   */
+  @Override
+  public PollableLocalTdbRdfStoreService create(final LocalTdbPoolKey key) throws Exception {
```
natan-cox-cognizone: soon to be fixed I assume.

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/jenamemory/LocalTdbRdfStoreService.java
```diff
@@ -1,317 +1,418 @@
...
+    }
 
-  private <T> T executeInLock(boolean lock, Supplier<T> executeInLock) {
-    model.enterCriticalSection(lock);
+    return isValid;
+  }
+
+  public QueryExecution createQueryExecution(final String query) {
+    final QueryExecution queryExecution = QueryExecutionFactory.create(query, dataset);
+    queryExecution.setTimeout(firstResultTimeout, firstResultTimeUnit, overallTimeout, overallTimeUnit);
```
natan-cox-cognizone: a bit weird. I guess all 3 methods can be delegated so we only need to setTimeout in 1 place?

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/jenamemory/LocalTdbRdfStoreService.java
```diff
@@ -1,317 +1,418 @@
...
+    if (log.isDebugEnabled()) {
+      log.debug("Running construct query: \n{}", query);
+    }
+    return callQueryConstruct(() -> createQueryExecution(query));
   }
 
-  @Override
-  public boolean graphExists(String graphUri) {
-    return datasetGraph.containsGraph(NodeFactory.createURI(graphUri));
+  public Model constructAllTriples() {
```
natan-cox-cognizone, line 286: Is this used? It seems to be a replacement for getModel. but the idea behind getModel might be flawed (except for in memory RdfStoreService... where it could be useful to get out the processed result.

natan-cox-cognizone: another option could be to have getModel do executeConstructQuery? or even copying the default model into a new model.

triplestore-tdb-pool/src/main/java/zone/cogni/asquare/triplestore/pool/factory/LocalTdbRdfStoreServiceFactory.java
gabor-harsanyi: yes-yes, it's pushed now :)

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/jenamemory/LocalTdbRdfStoreService.java
gabor-harsanyi: yes, good catch, thank you! it's pushed

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/jenamemory/LocalTdbRdfStoreService.java
gabor-harsanyi: Yes, I use `constructAllTriples` from the tests and also from the application, it was kind of needed in order to not rewrite whole parts of the application and all of these usages I wrapped around with a transaction.

> another option could be to have getModel do executeConstructQuery

I was thinking about that as well, but eventually decided to really discourage the usage of getModel with deprecation but still leaving an alternative option, this, in the documentation. I was not sure if on any application the model is used in any strange way already that I could break with changing it.

natan-cox-cognizone: ok, makes sense. thanks!

triplestore-jena-memory/src/test/java/zone/cogni/asquare/triplestore/jenamemory/LocalTdbRdfStoreServiceTest.java
```diff
@@ -0,0 +1,525 @@
...
+import org.apache.jena.query.QuerySolution;
+import org.apache.jena.query.ResultSet;
+import org.apache.jena.rdf.model.Literal;
+import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.ModelFactory;
+import org.apache.jena.rdf.model.RDFNode;
+import org.apache.jena.rdf.model.ResourceFactory;
+import org.apache.jena.sparql.engine.binding.Binding;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
```
natan-cox-cognizone, line 13: fyi I am quite a big fan of org.assertj.core.api.Assertions

natan-cox-cognizone (approved):

PR #61, opened 2022-02-08
move out pool as new module

gabor-harsanyi:
…) module


gabor-harsanyi:
This needs to be first approved and merged into https://github.com/cognizone/asquare/pull/60 PR if you want to have that fix in a bit more structured way

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/jenamemory/LocalTdbRdfStoreService.java
```diff
@@ -89,7 +85,7 @@ public LocalTdbRdfStoreService(
 
     this.dataset = TDBFactory.createDataset(tdbLocationFolder.getAbsolutePath());
     initialize(initFolder);
-    this.ready = true;
+    this.ready.set(true);
```
natan-cox-cognizone, line 88: if this is true when it does not fail, should we not have a finally for when it fails? maybe to do some cleanup related to what is setup during execution of the constructor?

triplestore-tdb-pool/src/main/java/zone/cogni/asquare/triplestore/pool/jenamemory/PollableLocalTdbRdfStoreService.java
```diff
@@ -0,0 +1,123 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import zone.cogni.asquare.triplestore.jenamemory.LocalTdbRdfStoreService;
+import zone.cogni.asquare.triplestore.pool.PoolableRdfStoreService;
+import zone.cogni.asquare.triplestore.pool.key.LocalTdbPoolKey;
+
+import java.io.File;
+import java.util.concurrent.TimeUnit;
+
+public class PollableLocalTdbRdfStoreService
```
natan-cox-cognizone, line 19: are we really having PollableLocalTdbRdfStoreService extend PoolableRdfStoreService? Seems like typo and if it is not it could be confusing?!

gabor-harsanyi: I'm not sure I understand you well, I think we do, but maybe I misunderstand you:
image

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/jenamemory/LocalTdbRdfStoreService.java
gabor-harsanyi: I didn't like the idea that we initialize TDB in the constructor but I don't want to change it because then we'd break backward compatibility.
My idea was to follow the original idea of the `LocalTdbRdfStoreService` where if a problem happens then it is not handled locally but propagated back on the caller chain, hence the `LocalTdbRdfStoreService` instance is not even constructed. If I'd introduce a new state where we handle issues locally and say may LocalTdbRdfStoreService instance exists but does not work well then we will need to re-work all the already existing codes to handle this new situation.
The `ready` is the same state as the `close` state was on the previous implementation, it's used the same way, just with opposite meaning

gabor-harsanyi: Cleanup would be nice tho' but I'm not sure if this is the best place to handle it because there can be a lot of issues and it's hard to now how to go back to a consistent state from here.

gabor-harsanyi: What I started to do is that I use LocalTdbRdfStoreService instances through the pool and when one instance creation fails then it is invalidated by the pool and on the application I listen for different type of issues during invalidation and see if we can react on them. For example if there is TDB corruption then we can react on it by disabling access to this TDB from other threads, put TDB into maintenance mode (this is a new mode on the app), export the content and re-create the TDB from export

triplestore-tdb-pool/src/main/java/zone/cogni/asquare/triplestore/pool/jenamemory/PollableLocalTdbRdfStoreService.java
gabor-harsanyi: My mistake, sorry, I couldn't see the `Poll` vs `Pool` typo :)

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/jenamemory/LocalTdbRdfStoreService.java
natan-cox-cognizone: ah, no. I fully agree that we should rethrow the issue. so, probably having a catch where we would cleanup resources (I dunno if there are any...) like connections, files etc could help. but indeed, construction should not be a "success" but should still throw an error.

so, probably not finally but catch with cleanup and rethrow.

gabor-harsanyi: My problem is that I cannot clean it up here, because one `LocalTdbRdfStoreService` instance only knows about its own state but knows not much about the global state of the TDB. The Dataset, or model or graph is just a local view of the global TDB whereas there can be more threads already working on this TDB. For example let's say I have an OOM thrown by this constructor but for that I don't need to delete the TDB because there might be other threads already/still working on it well.
I mean I could trick it around because from the Dataset view I can get the shared StoreConnection instance and from that I could get the shared transaction manager, then I could check if there is any active transactions, wait for the other threads to finish their jobs, commit the transactions and then clear up the things. But I considered this logic to be too complex to implement it here, I implemented this logic on top of the pool. So above this layer there is a pooling layer and above the pooling there is an error handling layer. (the last layer is only implemented on the application, because that is, I think, application dependent how it wants to handle different error cases on the TDB)

PR #67, opened 2022-03-29
added support for the "merged" operation files for the uris list flow

### LINKED ISSUE #68: operation file validation for list of uris does not cover the case of multiple operation files
The validate method in OperationResultProcessor that takes a list of uris should also handle the case where we have several operation files, similar as the method for single uri does. Additionally, we can improve the single uri method to accept suppliers

ssamarid:


cube/src/main/java/zone/cogni/asquare/cube/operation/OperationResultProcessor.java
```diff
@@ -176,18 +176,17 @@ private SingleGroupResult findValidateGroup(SingleGroupResult root, String curre
...
-                                         .map(uri -> {
-                                           SingleGroupResult rootGroup = new SingleGroupResult(operationGroup, uri, null);
-                                           processGroup(context, rootGroup);
-                                           return rootGroup;
-                                         })
-                                         .collect(Collectors.toList());
-
-    log.debug("(validate) {}", context.monitor);
-    return result;
+    try {
```
natan-cox-cognizone, line 188: yes this was probably never adapted for the multi-file solution, but in stead of duplicating code, here it probably makes sense have the method below this one call this method by delegation. also move the method below up (since it will be calling this one)

cube/src/main/java/zone/cogni/asquare/cube/operation/OperationResultProcessor.java
```diff
@@ -219,6 +218,13 @@ private SingleGroupResult processSingleFileCase(Context context, String uri, Str
     return result;
   }
 
+  private List<SingleGroupResult> processSingleFileCase(Context context, List<String> uris, String operationGroupId) {
```
natan-cox-cognizone: ah, I think my solution should work but maybe I missed something here. 

also, method should be moved up.

cube/src/main/java/zone/cogni/asquare/cube/operation/OperationResultProcessor.java
```diff
@@ -230,6 +236,13 @@ private SingleGroupResult processMergedFilesCase(Context context, String uri, St
     return rootGroupResult.getGroupResult(operationGroupId);
   }
 
+  private List<SingleGroupResult> processMergedFilesCase(Context context, List<String> uris, String operationGroupId) {
```
natan-cox-cognizone: move up

natan-cox-cognizone (changes requested):

natan-cox-cognizone (approved):

PR #70, opened 2022-04-04
remove zuul dependency

### LINKED ISSUE #69: remove zuul dependency
The zuul library is deprecated, so we should stop using it.

It is only used in one class which covers a simple use case.

natan-cox-cognizone:
remove SecureElasticLogFilter.java since it uses zuul; removed zuul from build file


alexey-lukashov (approved):

domivds (commented):
good job

Issue #99, opened 2022-07-21
Improve sorted RDF statements output in SortedStatementsString


natan-cox-cognizone:
add support for 

- base to shorten some URIs
- namespaces to create CURIEs
- configurable indentation
- `rdf:type` to be `a`

natan-cox-cognizone:
solved in branch #100

PR #100, opened 2022-07-21
add support for sorted ttl

natan-cox-cognizone:
- base to shorten some URIs
- namespaces to create CURIEs
- configurable indentation
- rdf:type to be a

### DIFF
--- a/cube/src/main/java/zone/cogni/asquare/cube/sort/SortedStatementsString.java
+++ b/cube/src/main/java/zone/cogni/asquare/cube/sort/SortedStatementsString.java
@@ -2,6 +2,7 @@ package zone.cogni.asquare.cube.sort;
 
 import org.apache.commons.collections4.MapUtils;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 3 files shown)


natan-cox-cognizone:
fiexes #99

cube/src/main/java/zone/cogni/asquare/cube/sort/SortedStatementsString.java
```diff
@@ -16,35 +17,98 @@
...
-    this.indent = indent;
+  public void setBase(String base) {
+    this.base = base;
+  }
+
+  public Map<String, String> getNamespaces() {
+    return namespaces;
+  }
+
+  public void setNamespaces(Map<String, String> namespaces) {
```
domivds, line 83: should we make a diff between setNamespaces and addNamespaces
this looks actually like add, not set

natan-cox-cognizone: hmm. true. this is only true in case of use of builder and we start with an empty map.

PR #107, opened 2022-07-26
add security2 module with saml2 (+introduce lombok)

domivds:


build.gradle
```diff
@@ -413,6 +418,22 @@ project('security') {
   }
 }
 
+project('security2') {
```
natan-cox-cognizone, line 421: naming by an engineer 😏

security2/src/main/java/zone/cogni/asquare/security2/saml2/Saml2Properties.java
```diff
@@ -0,0 +1,46 @@
...
+
+@Data
+public class Saml2Properties {
+
+  private SigningKeys signingKeyStore;
+  private String registrationId;
+  private String idpUrl;
+  private Attributes attributes;
+  private String roleMappingUrl;
+  private Map<String,User> basicAuthUsers = new HashMap<>(); //init so we allow empty config
```
natan-cox-cognizone:
```suggestion
  private Map<String, User> basicAuthUsers = new HashMap<>(); //init so we allow empty config
```

security2/src/main/java/zone/cogni/asquare/security2/saml2/Saml2HttpConfigurer.java
```diff
@@ -0,0 +1,84 @@
...
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+@RequiredArgsConstructor
+public class Saml2HttpConfigurer extends AbstractHttpConfigurer<Saml2HttpConfigurer, HttpSecurity> {
```
natan-cox-cognizone, line 33: sure this is only picked up if we enable saml?

natan-cox-cognizone (approved):

security2/src/main/java/zone/cogni/asquare/security2/saml2/Saml2HttpConfigurer.java
domivds: as long as this is not "bean-ized" somewhere we are ok

PR #130, opened 2022-08-18
Make sure local path prefix has / as separator

### LINKED ISSUE #129: IndexFolderService doesn't always use / as seperator of local path prefix
The method getLocalPathPrefix in IndexFolderService should always return a path with / as separator.
Whenever a path has a different separator, it should be replaced with /.

michielrogissart-cognizone:
fix issue #129

### DIFF
--- a/cube/src/main/java/zone/cogni/asquare/cube/index/IndexFolderService.java
+++ b/cube/src/main/java/zone/cogni/asquare/cube/index/IndexFolderService.java
@@ -96,15 +96,17 @@ public class IndexFolderService {
   }
 
   private String getLocalPathPrefix() {
-    String result = null;
+    String result;
     if (configurationPath.startsWith("classpath:"))
       result = StringUtils.substringAfter(configurationPath, "classpath:");
     else if (configurationPath.startsWith("file:"))
       result = StringUtils.substringAfter(configurationPath, "file:");
-    if (result != null) {
-      return File.separatorChar != '/' ? result.replace(File.separatorChar, '/') : result;
-    }
-    throw new RuntimeException("Unknown path prefix, should it be supported? path is '" + configurationPath + "'.");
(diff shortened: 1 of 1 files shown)


cube/src/main/java/zone/cogni/asquare/cube/index/IndexFolderService.java
```diff
@@ -96,11 +96,14 @@ private List<String> getLocalPaths(@Nonnull Resource[] resources) {
...
 
   private String getLocalPathPrefix() {
+    String result = null;
     if (configurationPath.startsWith("classpath:"))
-      return StringUtils.substringAfter(configurationPath, "classpath:");
+      result = StringUtils.substringAfter(configurationPath, "classpath:");
     else if (configurationPath.startsWith("file:"))
-      return StringUtils.substringAfter(configurationPath, "file:");
-
+      result = StringUtils.substringAfter(configurationPath, "file:");
```
natan-cox-cognizone, line 103: @michielrogissart-cognizone I would probably have an else with an exception here. it is more clear that it is supporting classpath/file or fail.

cube/src/main/java/zone/cogni/asquare/cube/index/IndexFolderService.java
```diff
@@ -96,11 +96,14 @@ private List<String> getLocalPaths(@Nonnull Resource[] resources) {
...
+    if (result != null) {
```
natan-cox-cognizone: this reminds me: normal flow should be free of if tests. exceptions should be in if-tests typically.

so that would mean more like if (result == null) throw exception.

natan-cox-cognizone (changes requested):

natan-cox-cognizone (approved):

Issue #141, opened 2022-08-30
bug: checking for missed triples and subjects won't log when 1 subject/triple is missed


michielrogissart-cognizone:

natan-cox-cognizone:
fixed by PR #142

PR #142, opened 2022-08-30
bug: 1 missed subject or triple not logged issue #141

michielrogissart-cognizone:


cube/src/main/java/zone/cogni/asquare/cube/convertor/ModelToJsonConversion.java
```diff
@@ -229,11 +229,10 @@ private void reportMissedSubjects(Context context, String root) {
   private void reportUnprocessedTriples(Context context, String root) {
     if (!log.isWarnEnabled()) return;
 
-    Model remainingModel = ModelFactory.createDefaultModel();
-    remainingModel.add(context.model);
-    remainingModel.remove(context.alreadyProcessedModel);
+    Model remainingModel = context.model.difference(context.alreadyProcessedModel);
+    remainingModel.setNsPrefixes(conversionProfile.getContext().getPrefixes());
```
natan-cox-cognizone: good idea but this can be a long list (currenly 15 long or so?)... and most of them would be unused. so either we drop this line... or you make sure you only log namespaces which are relevant to the model. 

in case you do the latter this might have a performance impact... so then I would also add an option in the configuration to enable or disable this. if disabled it would fall back to the default (i.e. not setting namespaces)

cube/src/main/java/zone/cogni/asquare/cube/sparql2json/SparqlSelectToJson.java
```diff
@@ -16,24 +17,82 @@
...
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
 public class SparqlSelectToJson {
 
-  private final List<Query> queries;
+  public static class QueryContext {
```
natan-cox-cognizone: huh?

natan-cox-cognizone (changes requested):

cube/src/main/java/zone/cogni/asquare/cube/convertor/ModelToJsonConversion.java
michielrogissart-cognizone: I'll change it to not add the prefixes then :)

cube/src/main/java/zone/cogni/asquare/cube/sparql2json/SparqlSelectToJson.java
michielrogissart-cognizone: I really don't know where this is coming from. Can this be from an older version that was not yet merged when I started the branch? I'll revert the changes.

natan-cox-cognizone (approved):

PR #146, opened 2022-09-01
fix #145 doing indexing in configurable blocks

### LINKED ISSUE #145: make indexing smarter so it can use less memory
Current `FolderBasedIndexingService` implementation create all `Callable` instances in go, which uses a lot of memory.

It would be better to be able to index in blocks of arbitrary size.

natan-cox-cognizone:


michielrogissart-cognizone (approved):

cube/src/main/java/zone/cogni/asquare/cube/index/StaticIndexIndexingService.java
```diff
@@ -70,14 +73,16 @@
    */
   private final Map<String, String> queryTemplateParameters;
 
-  public StaticIndexIndexingService(@Nonnull SpelService spelService,
+  public StaticIndexIndexingService(int indexBlockSize,
```
domivds, line 89: isn't this used yet? if so we need a (deprecated) constructor with a default value

cube/src/main/java/zone/cogni/asquare/cube/index/IndexFolderUriReport.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+class IndexFolderUriReport {
+
+  static class CollectionFolderUriReport {
+
+    private final IndexFolderUriReport indexFolderUriReport;
+    private final CollectionFolder collectionFolder;
+    private final List<String> uris;
+
+    public CollectionFolderUriReport(IndexFolderUriReport indexFolderUriReport,
```
domivds: if class is only used "interally", you can make it non-static and find parent class instance with `CollectionFolderUriReport.this`

cube/src/main/java/zone/cogni/asquare/cube/index/IndexFolderUriReport.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Keeps a list of all URIs linked to each of the collections. Only URIs are stored to save memory for indexing.
+ * When pulling out blocks of URIs which are going to be indexed the memory should remain at a predictable level.
+ */
+class IndexFolderUriReport {
+
+  static class CollectionFolderUriReport {
```
domivds: Inner classes at end of class

cube/src/main/java/zone/cogni/asquare/cube/index/IndexFolderUriReport.java
```diff
@@ -0,0 +1,125 @@
...
+      return collectionFolder;
+    }
+
+    public List<String> getUris() {
+      return Collections.unmodifiableList(uris);
+    }
+  }
+
+  private final IndexFolder indexFolder;
+  private final List<CollectionFolderUriReport> collectionFolderUriReports = new ArrayList<>();
```
domivds, line 14: not threadsafe (more a global comment, to be checked where needed as indexing runs threaded)

cube/src/main/java/zone/cogni/asquare/cube/index/IndexFolderUriReport.java
natan-cox-cognizone: true. still on the fence if I want to keep it in or move it out though.

cube/src/main/java/zone/cogni/asquare/cube/index/IndexFolderUriReport.java
natan-cox-cognizone: yes sir!

cube/src/main/java/zone/cogni/asquare/cube/index/IndexFolderUriReport.java
natan-cox-cognizone: this runs before we invoke the actual indexing into the threadpool

PR #147, opened 2022-09-07
fix: add namespace when namespace not present

### LINKED ISSUE #148: Shacl generator fails on an "unknown" namespace
make sure a prefix is automatically created

### LINKED ISSUE #149: improve indexing performance by using String and not Resource in CollectionFolder
so it can be used immediately as a string

michielrogissart-cognizone:


natan-cox-cognizone:
fix #148

cube/src/main/java/zone/cogni/asquare/cube/convertor/data2shacl/ShaclGenerator.java
```diff
@@ -172,8 +172,9 @@ private Resource calculateShapeBasedOnResource(@Nonnull Configuration configurat
 
     String namespacePrefix = shacl.getNsURIPrefix(originalResource.getNameSpace());
     if (namespacePrefix == null) {
-      throw new RuntimeException("no name alternative found for '" + originalResource.getURI() + "':"
-                                 + " please add namespace to prefixes.");
+      long nrTempPrefixes = shacl.getNsPrefixMap().keySet().stream().filter(key -> key.startsWith("ns")).count();
```
natan-cox-cognizone: hmm, no I think about it, this ok in most cases but what if there was already one or more "ns" prefixes present. the safer option is to find a good starting point and go from there.

roughly like 

```
private int getAvailableNamespaceIndex(...) {
  return shacl.getNsPrefixMap().keySet().stream()
            .filter(key -> key starts with)
            .map(ns -> ns.substring(2))
            .mapToInt(ns -> Integer.toInt(ns))
            .max();
} 
```

natan-cox-cognizone (changes requested):

cube/src/main/java/zone/cogni/asquare/cube/convertor/data2shacl/ShaclGenerator.java
```diff
@@ -172,15 +172,21 @@ private Resource calculateShapeBasedOnResource(@Nonnull Configuration configurat
...
 
     String prefixLocalName = firstPart == null ? namespacePrefix + "_" + localName
                                                : firstPart + "_" + namespacePrefix + "_" + localName;
     return ResourceFactory.createResource(configuration.getShapesNamespace() + prefixLocalName);
   }
 
+  private long getAvailableNamespaceIndex(Model shacl) {
+    return shacl.getNsPrefixMap().keySet().stream()
+            .filter(key-> key.matches("^ns[0123456789]*$"))
+            .count();
```
natan-cox-cognizone: Unfortunately, count is not a good way of calculating things. what if you have only ns2? @michielrogissart-cognizone

natan-cox-cognizone (approved):

PR #157, opened 2022-09-30
add a difference method for models which works for all typed Literals

### LINKED ISSUE #156: Jena Model.difference method does not work with all typed Literal
Propose to add a method in JenaUtils which does a difference correctly.

natan-cox-cognizone:
fix #156

### DIFF
--- a/cogni-libs/src/main/java/zone/cogni/libs/jena/utils/JenaUtils.java
+++ b/cogni-libs/src/main/java/zone/cogni/libs/jena/utils/JenaUtils.java
@@ -586,27 +586,38 @@ public class JenaUtils {
 //  }
 
 
+  /**
+   * Because of a bug in Jena Model difference does not work on Literal with special datatypes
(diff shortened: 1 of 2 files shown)


cogni-libs/src/main/java/zone/cogni/libs/jena/utils/JenaUtils.java
```diff
@@ -584,5 +585,54 @@ public static Literal createTypedLiteral(Model model, String literal, String dat
...
+            if (!object.isLiteral()) {
+              actualResult.add(statement);
+            }
+
+            Literal literal = object.asLiteral();
+            if (literal.getDatatype() == null) {
+              actualResult.add(statement);
+            }
+
+            if (!hasMatchingObject(b, statement)) {
```
domivds: maybe add a comment why we do this, i know we discussed this, but other people will be clueless

domivds (approved):

PR #161, opened 2022-10-11
operation template with more context

### LINKED ISSUE #160: operations need to have a variable context passed in the calculation
Some operations need to have a context that is variable between calls, e.g. when evaluating something based on the security role of the current user

ssamarid:
- operation templates can take an additional set of context params
- builder can handle several cases of input

### DIFF
--- a/cube/src/main/java/zone/cogni/asquare/cube/operation/OperationResultJsonConversion.java
+++ b/cube/src/main/java/zone/cogni/asquare/cube/operation/OperationResultJsonConversion.java
@@ -7,14 +7,14 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.jena.rdf.model.Model;
 import org.apache.jena.rdf.model.ModelFactory;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import zone.cogni.asquare.cube.operation.OperationResultProcessor.OperationValidationContext;
 import zone.cogni.asquare.cube.operation.OperationResultProcessor.SingleGroupResult;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


cube/src/main/java/zone/cogni/asquare/cube/operation/OperationResultJsonConversion.java
```diff
@@ -42,6 +42,16 @@ public JsonNode createStandaloneJson(Supplier<Set<String>> permissionsSupplier,
     return arrayNode;
   }
 
+  public JsonNode createStandaloneJson(OperationValidationContext context,
+                                       List<String> uris) {
+    ArrayNode arrayNode = JsonNodeFactory.instance.arrayNode();
+
+    List<SingleGroupResult> results = uris.stream().map(uri-> operationResultProcessor.validate(context, uri)).collect(Collectors.toList());
+    results.forEach(result -> arrayNode.add(createGroupJson(result)));
```
natan-cox-cognizone, line 50: @ssamarid I think we have an addAll method on arrayNode

cube/src/main/java/zone/cogni/asquare/cube/operation/OperationResultJsonConversion.java
```diff
@@ -50,13 +60,23 @@ public ObjectNode createStandaloneJson(Supplier<Set<String>> permissionsSupplier
     return createStandaloneJson(permissionsSupplier, () -> model, uri, operationGroupId);
   }
 
+  public ObjectNode createStandaloneJson(OperationValidationContext context) {
+    String uri = StringUtils.EMPTY;
```
natan-cox-cognizone, line 64: should uri be empty, not null?

natan-cox-cognizone (approved):

PR #163, opened 2022-10-18
fix on dateTime values during model to json conversion

### LINKED ISSUE #162: ModelToJsonConversion should not modify values
In the case of xsd:dateTime, the value is parsed and then formatted into ISO, potentially changing the value if the original did not have the same format. It is a good idea to keep the parsing in order to ensure we have valid dateTime formats in the data, but the conversion should just keep the value as it receives it.

ssamarid:


cube/src/test/java/zone/cogni/asquare/cube/convertor/ModelToJsonConversionTest.java
```diff
@@ -291,6 +292,32 @@ public void context_test() {
...
+
+  @Test
+  public void datetime_parse_not_ok() {
+    // given
+    Model homerModel = JenaUtils.read(new ClassPathResource("convertor/person-data-homer-datetime-parse-bad-format.ttl"));
+    ModelToJsonConversion conversion = getContextPersonConversion(getContextConfiguration());
+
+    // when
+    // then
+    assertThatExceptionOfType(RuntimeException.class).isThrownBy(()-> conversion.apply(homerModel, "http://demo.com/data#homer"));
```
natan-cox-cognizone: you can also check part of the contents of the exception. just to make the test a bit clearer... and to avoid any random exception is making the test pass. use `.withMessageContaining`

natan-cox-cognizone (approved):

PR #168, opened 2022-10-25
166 go back to jdk 8

### LINKED ISSUE #166: Go back to JDK 8
Some projects are still on Java 8 and it would be nice to be able to support those too with the latest versions.

### LINKED ISSUE #167: Improve asquare:cube module dependencies
Some spring dependencies are exposed as "api" dependencies in gradle which is not needed.

Investigate if we can switch to "compile".

natan-cox-cognizone:


natan-cox-cognizone:
fix #166

natan-cox-cognizone:
fixes #167

build.gradle
```diff
@@ -111,8 +112,8 @@ subprojects {
     compileOnly 'org.projectlombok:lombok'
   }
 
-  sourceCompatibility = JavaVersion.VERSION_11
-  targetCompatibility = JavaVersion.VERSION_11
+  sourceCompatibility = "1.8"
```
gabor-harsanyi, line 115: no need to change, I just put it here if you want to follow the same way of configuring like java 11, then it is JavaVersion.VERSION_1_8

gabor-harsanyi (approved):

ssamarid (approved):

PR #172, opened 2022-11-30
feat: T15052(hanami) model hash computation

michielrogissart-cognizone:


cube/src/main/java/zone/cogni/asquare/cube/sort/StringRdfVisitor.java
```diff
@@ -49,6 +52,40 @@ public String visitLiteral(Literal literal) {
...
+  @Override
+  public String visitStmt(Resource r, Statement statement) {
+    String result = "";
+
+    Resource subject = statement.getSubject();
+    Property predicate = statement.getPredicate();
+    RDFNode object = statement.getObject();
+
+    // visit subject
+    if (subject.isAnon()) {
```
natan-cox-cognizone: you do not need this. the idea of a visitor is to do something like

`String result = (String) rdfNode.visitWith(StringRdfVisitor.instance);`

so visitor takes care of all types automagically.

cube/src/main/java/zone/cogni/asquare/cube/index/IndexMethod.java
```diff
@@ -211,11 +213,18 @@ private ObjectNode addFacets(Model draftModel, ObjectNode objectNode, String uri
     Map<String, RDFNode> bindings = ImmutableMap.of("uri", ResourceFactory.createResource(uri));
 
     ObjectNode facetNode = sparqlSelectToJson.convert(draftModel, bindings);
+    facetNode.put("hash", getHash(draftModel));
```
natan-cox-cognizone: this is pretty drastic. it would imply that all apps get this. whether they want it or not.

cube/src/main/java/zone/cogni/asquare/cube/hash/ModelHasher.java
```diff
@@ -0,0 +1,70 @@
...
+        return stmt.getSubject().isAnon() || stmt.getObject().isAnon();
+    }
+
+    private static byte[] hashFromSortedTurtle(Model model) {
+        SortedStatementsString sortedStatementsString = new SortedStatementsString();
+        String sortedString = sortedStatementsString.apply(model);
+
+        try {
+            MessageDigest sha256 = MessageDigest.getInstance("SHA-256");
+            byte[] bytes = sortedString.trim().getBytes(StandardCharsets.UTF_8);
```
natan-cox-cognizone: let's not trim... let's keep as is.

cube/src/main/java/zone/cogni/asquare/cube/index/IndexMethod.java
natan-cox-cognizone: ok, I checked. there is a `Configuration` class. we can use this to add a flag for adding hashes.

natan-cox-cognizone (changes requested):

cube/src/main/java/zone/cogni/asquare/cube/hash/ModelHasher.java
michielrogissart-cognizone: I saw that the sortedString started with a newline, that's why I added it.

natan-cox-cognizone: let's fix the sorted string then :)

natan-cox-cognizone (approved):

PR #174, opened 2022-12-20
feat: T15398 (hanami) ensure index exists

michielrogissart-cognizone:


cube/src/main/java/zone/cogni/asquare/cube/index/IndexingServiceContext.java
```diff
@@ -24,4 +31,31 @@ abstract class IndexingServiceContext {
 
   protected abstract Map<String, String> getQueryTemplateParameters();
 
+  protected ElasticsearchMetadataService getElasticsearchMetadataService() {
+    return elasticsearchMetadataService;
+  }
+
+  @Override
+  public void ensureIndexExists(@Nonnull String index) {
```
natan-cox-cognizone: ok, I got it. I use `InternalIndexingServiceUtils` because I do not like to put too much code in superclasses. This often becomes a pain if there are too many exceptions in subclasses.

Hence my design where I kept the classes for the most part separate and just delegate to the (badly named) Utils class

natan-cox-cognizone (approved):

PR #198, opened 2023-03-16
switch to Jena SHACL class; also fixed some other deprecations

### LINKED ISSUE #197: Use Jena for SHACL ontology
and not our own code.

natan-cox-cognizone:


service/src/main/java/zone/cogni/asquare/service/rest/model/RestResponse.java
```diff
@@ -44,10 +44,7 @@ public String getTitle() {
   }
 
   public String getCode() {
-    if (StringUtils.isEmpty(code)) {
-      return "";
-    }
-    return code;
+    return StringUtils.hasText(code) ? code : "";
```
domivds: not sure if we should enforce usage of apache commons lang3 if needed function is available?

michielrogissart-cognizone (approved):

domivds (approved):

PR #204, opened 2023-04-28
Datetime validation: allow values without timezone

domivds:


cube/src/main/java/zone/cogni/asquare/cube/convertor/ModelToJsonConversion.java
```diff
@@ -813,7 +814,12 @@ private String literalToDate(Literal literal) {
   private String literalToDateTime(Literal literal) {
     String stringValue = literal.getLexicalForm();
     // we do not do anything with this, it just validates that the dateTime value is in a parsable format
-    ZonedDateTime.parse(stringValue);
+    try {
+      ZonedDateTime.parse(stringValue);
+    }
+    catch (Exception e) {
```
natan-cox-cognizone, line 820: do we need to catch a specific exception?

natan-cox-cognizone (approved):

ssamarid (approved):

PR #208, opened 2023-06-19
feat: configurable SparqlSelectToJson unexpected list handling

michielrogissart-cognizone:


natan-cox-cognizone (approved):

cube/src/main/java/zone/cogni/asquare/cube/sparql2json/SparqlSelectToJson.java
```diff
@@ -137,6 +148,27 @@ else if (!isPresent && propertyConversion.isList()) {
     }
   }
 
+  private void handleListException(ObjectNode current, String lastLevel, JsonNode convertedValue, PropertyConversion propertyConversion) {
+    switch(listExceptionHandling) {
+      case fail:
+        throw new RuntimeException("multiple results found for " + propertyConversion.getPropertyName());
+      case ignoreAll:
```
domivds: isn't "ignoreAll" confusing? You get here when 1 is already added, so maybe something like "takeFirstInput" and rename "takeFirst" to "takeFirstSorted" ?  
(takeFirstInput makes it clear that you can sort in your query)

domivds (approved):

PR #220, opened 2023-10-09
#219 Allow to do select query and handle the result row by row (per QuerySolution)

### LINKED ISSUE #219: Add paginated select query with direct "result row" handling
If the caller can handle each row 1 by 1, we can avoid creating intermediate list objects

domivds:
important change is about the PaginatedQuery

One approved the change will be cherry-picked to develop


natan-cox-cognizone (approved):

cube/src/main/java/zone/cogni/asquare/cube/pagination/PaginatedQuery.java
```diff
@@ -152,6 +154,39 @@ private Model getModel(RdfStoreService rdfStore, String constructQuery, long off
...
+      int queryReturnedSize = selectBatch(rdfStore, limitQuery, querySolutionConsumer);
+      log.info("Query offset {} and limit {} returned {} rows", offset, batchSize, queryReturnedSize);
+      if (queryReturnedSize < batchSize) return;
+      offset += batchSize;
+    }
+  }
+
+  private int selectBatch(RdfStoreService rdfStore, String query, Consumer<QuerySolution> querySolutionConsumer) {
+    return rdfStore.executeSelectQuery(query, resultSet -> {
+      int counter = 0;
```
ssamarid, line 208: maybe stream first instead of hasNext()/next()? something like:
Streams.stream(resultSet).peek(bla -> querySolutionConsumer.accept(bla)).count()?

domivds: yes but no, google guava Streams.stream is annotated as beta

ssamarid (approved):

Issue #221, opened 2023-10-23
Add skolemization support


natan-cox-cognizone:
Add support to skolemize blank nodes. For this we can use functionality part of statement sorting where we can use the hashes to generate the skolem URIs.

PR #222, opened 2023-10-23
#221 added skolem support

natan-cox-cognizone:


cube/src/main/java/zone/cogni/asquare/cube/sort/SkolemizedModel.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.List;
+import java.util.function.Function;
+
+public class SkolemizedModel implements Function<Model, Model> {
+
+  public SkolemizedModel(@Nonnull String skolemBaseUri) {
+    this.skolemBaseUri = skolemBaseUri;
+  }
+
+  private String skolemBaseUri;
```
michielrogissart-cognizone: final? maybe also above the constructor?

cube/src/main/java/zone/cogni/asquare/cube/sort/SkolemizedModel.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Override
+  public Model apply(Model model) {
+    Model result = ModelFactory.createDefaultModel();
+    List<Statement> statements = new StatementSorter().apply(model);
```
michielrogissart-cognizone, line 26: Do the statements need to be sorted? Might give some overhead on big models

cube/src/main/java/zone/cogni/asquare/cube/sort/SkolemizedModel.java
```diff
@@ -0,0 +1,48 @@
...
+
+    statements.forEach(statement -> {
+      Resource newSubject = statement.getSubject().isURIResource() ? statement.getSubject()
+                                                                   : getSkolemUri(statement.getSubject());
+
+      RDFNode newObject = !statement.getObject().isAnon() ? statement.getObject()
```
michielrogissart-cognizone, line 32: I remember you not being a fan of ! operator if not needed. I guess it makes sense to align with other check, so yeah, it's the same for me :)

cube/src/main/java/zone/cogni/asquare/cube/sort/SkolemizedModel.java
```diff
@@ -0,0 +1,48 @@
...
+      result.add(newStatement);
+    });
+
+    return result;
+  }
+
+  private Resource getSkolemUri(RDFNode blankNode) {
+    return ResourceFactory.createResource(
+            StringUtils.appendIfMissing(skolemBaseUri, "/")
+                    + ".well-known/genid/" + blankNode.asResource().getId().getLabelString()
```
michielrogissart-cognizone, line 45: Maybe the StringUtils call can go in constructor? Would make this method a little bit more readable (but not a big problem obviously)
Also, I thought you were going to use the hash of the blank node, but this is fine as well for me. :)

cube/src/main/java/zone/cogni/asquare/cube/sort/SkolemizedModel.java
natan-cox-cognizone: for now: yes. it will be much more work to make it work otherwise.

cube/src/main/java/zone/cogni/asquare/cube/sort/SkolemizedModel.java
natan-cox-cognizone: yes. neither am I but else the expression would be much more complex. we need then to check for URI resource or literal.

cube/src/main/java/zone/cogni/asquare/cube/sort/SkolemizedModel.java
natan-cox-cognizone: blank node id here is same as the hash btw

michielrogissart-cognizone (approved):

natan-cox-cognizone:
fix #221

PR #225, opened 2023-11-16
T18500 catch TemplateEngineException

michielrogissart-cognizone:


elasticproxy/src/main/java/zone/cogni/asquare/elasticproxy/ElasticsearchProxyTemplate.java
```diff
@@ -225,10 +261,16 @@ private String spel(String expression, Map<String, Object> map) {
...
+  @Nullable
   private String thymeleaf(String expression, Map<String, Object> map) {
     Context ctx = new Context();
     ctx.setVariables(map);
-    return templateEngine.process(expression, ctx);
+    try {
+      return templateEngine.process(expression, ctx);
+    }
+    catch (TemplateEngineException e) {
+      return null;
```
natan-cox-cognizone: I usually do not do this but fine.

natan-cox-cognizone (approved):

elasticproxy/src/main/java/zone/cogni/asquare/elasticproxy/ElasticsearchProxyTemplate.java
```diff
@@ -191,7 +191,16 @@ private ResponseEntity<Resource> executeRequest(String requestUUID, HttpMethod h
     Request request = new Request(httpMethod.name(), buildUrl(spel(url, params), urlParams));
 
     if (HttpMethod.POST.equals(httpMethod) || HttpMethod.PUT.equals(httpMethod)) {
-      request.setJsonEntity(thymeleaf(getTemplateContent(), params));
+      String templatedBody = thymeleaf(getTemplateContent(), params);
+
+      if (templatedBody == null) {
```
ssamarid: what if we skip the null and instead catch the exception here, rather than in the other method? then this logic can be in the catch here 🤷 

And then it could be in the same try/catch as the one below. And since it looks like we need a separate treatment for this one (compared to the existing that will put the thymeleaf error message in the response), have 2 catch clauses?

michielrogissart-cognizone: I actually implemented it like this at first but then changed it because I felt that the exception should be catched asap. I'll change it :)

elasticproxy/src/main/java/zone/cogni/asquare/elasticproxy/ElasticsearchProxyTemplate.java
```diff
@@ -225,9 +258,11 @@ private String spel(String expression, Map<String, Object> map) {
     return expressionParser.parseExpression(expression, new TemplateParserContext()).getValue(context, String.class);
   }
 
+  @Nullable
```
ssamarid: I guess this can go?

elasticproxy/src/main/java/zone/cogni/asquare/elasticproxy/ElasticsearchProxyTemplate.java
```diff
@@ -20,18 +20,18 @@
...
 import org.thymeleaf.templateresolver.StringTemplateResolver;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
ssamarid: you may want to change your settings about this one

ssamarid (approved):

elasticproxy/src/main/java/zone/cogni/asquare/elasticproxy/ElasticsearchProxyTemplate.java
```diff
@@ -212,6 +213,38 @@ private ResponseEntity<Resource> executeRequest(String requestUUID, HttpMethod h
                 response.getStatusLine());
       return ResponseEntity.badRequest().body(new InputStreamResource(IOUtils.toInputStream(requestUUID, StandardCharsets.UTF_8)));
     }
+    catch (TemplateEngineException e) {
+      String stringifiedParameters = stringifyParameters(params);
+      String errorMessage = "Processing of template '" + getName() + "' with parameters " + stringifiedParameters + " failed.";
+      return ResponseEntity.status(500).body(new InputStreamResource(new ByteArrayInputStream(errorMessage.getBytes())));
```
ssamarid, line 224: can you also log this?

PR #229, opened 2024-01-18
feat: #23 Support for Jena 4

### LINKED ISSUE #23: Investigate upgrade of Jena
Currently Jena is on v 3.10.0. Latest 3.x.x is release is 3.17.0.

psiotwo:
Notes
- upgrade Java 8 -> Java 11
- upgrade Jena 3.17.0 -> Jena 4.10.0
- upgrade Apache HttpClient -> JDK11 HttpClient
- upgrade SpringBoot 2.6.6 -> 2.7.18
- upgrade SpringSecurity 5.6.2 - > 5.8.9
- upgrade SnakeYaml 1.29 -> 2.0
- unified response handling for `SparqlService` implementations
- backing SparqlService implementations by Jena 4/Java HttpClient, backing Virtuoso service with `RdfConnection`
- added a few tests for the Fuseki service as a representative for others.
- added (disabled) integration tests for GraphDB and Virtuoso

@natan-cox-cognizone @domivds a few questions:
- https://github.com/cognizone/asquare/blob/aade719644a13445069879c646711629ce0dfcd3/triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/sparqlendpoint/SparqlEndpointDatabase.java#L77 - above in the file it is mentioned that this is probably not used. Any hints what should be the fate of this switch? A similar one is also in `SparqlEndpointRdfStoreService.java`. 
- https://github.com/cognizone/asquare/blob/aade719644a13445069879c646711629ce0dfcd3/cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/GraphDBSparqlService.java#L41 - not sure for what it was used? For grabbing the proxy configuration from system props? Anyway, currently it is commented out.
- https://github.com/cognizone/asquare/blob/aade719644a13445069879c646711629ce0dfcd3/cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/VirtuosoSparqlService.java#L97 - again not sure, what was the use-case for turning off caching explicitly.

Outstanding:
- testing against Stardog


access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
```diff
@@ -60,20 +49,18 @@ public VirtuosoRdfStoreService(String rdfStoreUrl, String rdfStoreUser, String r
     this.rdfStoreUser = rdfStoreUser;
     this.rdfStorePassword = rdfStorePassword;
     this.graphCrudUseBasicAuth = graphCrudUseBasicAuth;
+    //  TODO HttpClientBuilder httpClientBuilder = HttpClients.custom().useSystemProperties();
+    httpClient = Utils.createHttpClientBuilder(rdfStoreUser, rdfStorePassword).connectTimeout(
```
natan-cox-cognizone: seems we set password here and also in httpClientBuilder method. it might be that this is how it should be done. but if is how it should be done: then please add a comment

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/Utils.java
```diff
@@ -0,0 +1,89 @@
...
+   * @return HttpClient.Builder
+   */
+  public static HttpClient.Builder createHttpClientBuilder(final String username,
+      final String password) {
+    final HttpClient.Builder httpClientBuilder = HttpClient.newBuilder();
+
+    if (StringUtils.isNoneBlank(username, password)) {
+      httpClientBuilder.authenticator(createBasicAuthenticator(username, password));
+    } else {
+      log.error("Endpoint credentials not properly configured");
```
natan-cox-cognizone: a little soft maybe. depends on what is checked before `createHttpClientBuilder` is called also.

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/Utils.java
```diff
@@ -0,0 +1,89 @@
...
+   * Executes a request and checks its response for success.
+   *
+   * @param request HttpRequest to analyze.
+   * @param client HttpClient to use.
+   */
+  public static void execute(final HttpRequest request, final HttpClient client) {
+    try {
+      HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+      checkOK(response);
+    } catch (IOException | InterruptedException e) {
```
natan-cox-cognizone: definitely not an `InterruptedException` expert. but should we typically not throw an `InterruptedException` in this thread... so all thread can shutdown?

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/Utils.java
```diff
@@ -0,0 +1,89 @@
...
+import java.net.Authenticator;
+import java.net.PasswordAuthentication;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+
+/**
+ * Utility methods for JDK11 HttpClient.
+ */
+public class Utils {
```
natan-cox-cognizone: HttpClientUtils?

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
```diff
@@ -128,67 +94,37 @@ public <R> R executeSelectQuery(Query query, QuerySolutionMap bindings, JenaResu
...
   public Model executeConstructQuery(Query query, QuerySolutionMap bindings) {
     query = buildQuery(query, bindings);
-
-    try (CloseableHttpClient httpClient = buildHttpClient(); QueryEngineHTTP queryExecution = (QueryEngineHTTP) QueryExecutionFactory.sparqlService(rdfStoreUrl, query,
-                                                                                                                                                    httpClient)) {
-      if (modelContentType != null) {
-        queryExecution.setModelContentType(modelContentType);
-      }
+    try (QueryExecution queryExecution = queryExecutionBuilder.query(query).build()) {
+      // TODO if (modelContentType != null) { queryExecution.setModelContentType(modelContentType); }
```
natan-cox-cognizone: @domivds please help Petr. I guess you know what this is about.

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
```diff
@@ -51,6 +36,10 @@ public class VirtuosoRdfStoreService implements RdfStoreService {
   private final boolean graphCrudUseBasicAuth;
   protected String modelContentType;
 
+  private HttpClient httpClient;
+
+  private QueryExecutionBuilder queryExecutionBuilder;
```
natan-cox-cognizone: is this builder threadsafe? assuming we do not update its properties.

cube/src/test/java/zone/cogni/asquare/cube/urigenerator/UriGeneratorCalculatorTest.java
```diff
@@ -26,6 +25,7 @@ public void test_uris_converted() {
     //when
     Model converted = uriCalculator.convert(model, ImmutableMap.of("baseUri", "http://asquare.cogni.zone"));
     //then
+    converted.write(System.out, Lang.TURTLE.getName());
```
natan-cox-cognizone: little strange. you did not even change any code for URI generator. 😃

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/sparqlendpoint/SparqlEndpointDatabase.java
natan-cox-cognizone, line 1: Trying out github code search and it does not find it being used at all. Except for this PR.
Mainly a note "to self" to start thinking about removing code.

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/sparqlendpoint/SparqlEndpointRdfStoreService.java
natan-cox-cognizone, line 1: Similarly, this class is (I guess) only used in ESCO CTMS. We could move it there.
Again, more a not to self.

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/JenaModelSparqlService.java
```diff
@@ -58,7 +59,7 @@ public void upload(Model model, String graphUri) {
 
     private Dataset getDatasetForSelect() {
         if (simulateRelaxedVirtuosoSparqlSelect) {
-            Dataset relaxedDataset = DatasetFactory.create(dataset);
+            Dataset relaxedDataset = DatasetFactory.wrap(DatasetGraphFactory.cloneStructure(dataset.asDatasetGraph()));
```
natan-cox-cognizone, line 62: why do we need to clone the DatasetGraph?

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
domivds: doesn't ring a bell
i would disregard this, modelContentType has only protected access and I see no usages

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
psiotwo: Good catch - probably some relict from refactoring ... fixed.

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/Utils.java
psiotwo: Hmm, probably yes, wasn't thinking much about it ... yet, this would require propagating up even through the SparqlService API ... Another options is an async call with `.join` ending in unchecked exceptions only - I updated the code according to this second option - feel free to comment.

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/Utils.java
psiotwo: This logic tries to unify different behaviour for username/password that was exhibited by the different backend implementations before (different conditions on username/password being (not)null). Currently:
- both username and password present means authentication should be used,
- either username and password not present means authentication should not be used.
In this sense, it is quite simple.

An alternative:
- both username/password is present -> authentication should be used,
- none of username/password is present -> authentication should not be used,
- either username/password is not present -> an exception is thrown.

cube/src/test/java/zone/cogni/asquare/cube/urigenerator/UriGeneratorCalculatorTest.java
psiotwo: Indeed ... again some relict of testing probably. Thx for noticing. Reverted (except of the cleaned up imports).

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/JenaModelSparqlService.java
psiotwo: Well, I only rewrote the original logic ... Personally, I don't see any reason for cloning it either, but maybe @natan-cox-cognizone [knows someone who knows](https://github.com/cognizone/asquare/blame/554c0a697cf172f62fa194dc1dd0ec1da8a20b8e/cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/JenaModelSparqlService.java#L61) about some :-D ...

```
 public static Dataset create(Dataset dataset) {
    Objects.requireNonNull(dataset, "Clone dataset is null");
    return new DatasetImpl(dataset);
  }
```
...
```
public DatasetImpl(Dataset ds) {
    this(DatasetGraphFactory.cloneStructure(ds.asDatasetGraph()));
}
  ```

File: access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java (hunk 6 above)
Comment (psiotwo) [reply access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java]: Good catch - checked the implementation and `QueryExecutionBuilderAdapter` mutates its content, so extracted it to the builder methods.

File: access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java [hunk 9]
```diff
@@ -60,20 +46,13 @@ public VirtuosoRdfStoreService(String rdfStoreUrl, String rdfStoreUser, String r
     this.rdfStoreUser = rdfStoreUser;
     this.rdfStorePassword = rdfStorePassword;
     this.graphCrudUseBasicAuth = graphCrudUseBasicAuth;
+    //  TODO HttpClientBuilder httpClientBuilder = HttpClients.custom().useSystemProperties();
```
domivds: this is mainly needed to configure a proxy (but also takes up other stuff like keepalive)
the "system properties builder" takes lot of configuration from the system properties as default
not sure where the java httpclient takes the default values from, might work out of the box, to test

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
```diff
@@ -60,20 +46,13 @@ public VirtuosoRdfStoreService(String rdfStoreUrl, String rdfStoreUser, String r
...
-      CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
-      credentialsProvider.setCredentials(AuthScope.ANY, new UsernamePasswordCredentials(rdfStoreUser, rdfStorePassword));
-      httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider);
-    }
-    else {
-      log.warn("Virtuoso executor service {} is configured without credentials.", rdfStoreUrl);
-    }
-    return httpClientBuilder.build();
+  private HttpClient.Builder httpClientBuilder(final boolean withAuthentication) {
+    return HttpClientUtils.createHttpClientBuilder(withAuthentication ? rdfStoreUser : null, rdfStorePassword).connectTimeout(Duration.ofSeconds(60));
```
domivds: you need to make a difference between graphCrudUseBasicAuth enabled or not
if not, virtuoso will by default use digest, if enabled you need to use basic-auth and directly pass the Authorization header

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
```diff
@@ -128,67 +86,36 @@ public <R> R executeSelectQuery(Query query, QuerySolutionMap bindings, JenaResu
...
-      }
-    }
-    catch (IOException ex) {
-      log.error("Virtuoso update failed {}", ex.getMessage());
-      throw new VirtuosoOperationException();
-    }
+    final HttpRequest request = HttpRequest
+        .newBuilder(URI.create(rdfStoreUrl))
+        .POST(BodyPublishers.ofString("query=" + updateQuery, StandardCharsets.UTF_8))
+        .header(CONTENT_TYPE, Lang.TURTLE.getHeaderString())
```
domivds: turtle?

build.gradle
```diff
@@ -99,12 +103,14 @@ subprojects {
...
       dependency("org.springframework.boot:spring-boot-starter-web:$springBootVersion")
-      dependency("org.springframework.security.extensions:spring-security-saml2-core:1.0.10.RELEASE")
+      dependency("org.springframework.security:spring-security-core:$springSecurityVersion")
+      dependency("org.springframework.security.extensions:spring-security-saml2-core:$springSecuritySaml2CoreVersion")
 
       // test
       dependency("ognl:ognl:3.1.12") // seems thymeleaf needs this library
       dependency("org.elasticsearch.test:framework:$elasticsearchVersion")
       dependency("org.springframework.boot:spring-boot-starter-test:$springBootVersion")
+      dependency("org.apache.jena:jena-fuseki-main:$jenaVersion")
```
domivds, line 114: if test as in dependencies for test, shouldn't this be `testImplementation` instead of `dependency` (or something alike)

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/FusekiSparqlService.java
```diff
@@ -39,59 +37,43 @@ public FusekiSparqlService(Config config) {
...
-              .body(new UrlEncodedFormEntity(Collections.singletonList(new BasicNameValuePair("update", updateQuery)), StandardCharsets.UTF_8))
-              .execute();
-      checkAndDiscardResponse(response);
-    }
-    catch (IOException e) {
-      throw new RuntimeException(e);
-    }
+      final HttpRequest request = HttpRequest
+          .newBuilder(URI.create(config.getUpdateUrl()))
+          .POST(BodyPublishers.ofString("update="+ updateQuery, StandardCharsets.UTF_8))
```
domivds: is this ok? i think form body is of format key=encodedValue
here the = will also be encoded

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/GraphDBSparqlService.java
```diff
@@ -1,72 +1,48 @@
...
-      CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
-      credentialsProvider.setCredentials(AuthScope.ANY, new UsernamePasswordCredentials(config.getUser(), config.getPassword()));
-      httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider);
-    }
-    else if (!StringUtils.isAllBlank(config.getUser(), config.getPassword())) {
-      log.error("Endpoint credentials not properly configured");
-    }
-    this.httpClient = httpClientBuilder.build();
+    //  TODO check loading from systemproperties - e.g. proxy settings?
+    //  HttpClientBuilder httpClientBuilder = HttpClients.custom().useSystemProperties();
```
domivds, line 35: see previous comment

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/GraphDBSparqlService.java
```diff
@@ -1,72 +1,48 @@
...
+    httpClient = HttpClientUtils.createHttpClientBuilder(config.getUser(), config.getPassword()).build();
+    queryExecutionBuilder = QueryExecutionHTTPBuilder.service(config.getSparqlEndpoint()).httpClient(httpClient);
   }
 
   @Override
   public void uploadTtlFile(File file) {
     try {
-      uploadTtl(file.getName(), file.toURI().toString(), FileUtils.readFileToString(file, "UTF-8"));
+      uploadTtl(file.toURI().toString(), FileUtils.readFileToString(file, "UTF-8"));
```
domivds, line 46: can't we use StandardCharsets.UTF_8 now?

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/GraphDBSparqlService.java
```diff
@@ -77,52 +53,46 @@ public void uploadTtlFile(File file) {
...
-    }
-    catch (JsonProcessingException e) {
-      throw new RuntimeException(e);
-    }
+  private void uploadTtl(String graphUri, String ttl) {
+    final HttpRequest request = HttpRequest
+        .newBuilder()
+        .POST(BodyPublishers.ofString(ttl))
+        .header(CONTENT_TYPE, Lang.TURTLE.getHeaderString())
+        .uri(URI.create(config.getSparqlUpdateEndpoint()+"?context=" + URLEncoder.encode("<"+graphUri+">", Charset.defaultCharset())))
```
domivds: shouldn't this be UTF-8 instead of defaultCharset?

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/GraphDBSparqlService.java
```diff
@@ -77,52 +53,46 @@ public void uploadTtlFile(File file) {
...
 
   @Override
   public void executeUpdateQuery(String updateQuery) {
-    HttpPost request = new HttpPost(config.getSparqlUpdateEndpoint());
-    request.setEntity(new UrlEncodedFormEntity(Collections.singletonList(new BasicNameValuePair("update", updateQuery)), StandardCharsets.UTF_8));
-
-    executeHttpRequest(request, 204);
+    final HttpRequest request = HttpRequest
+        .newBuilder(URI.create(config.getSparqlUpdateEndpoint()))
+        .POST(HttpRequest.BodyPublishers.ofString("update=" + updateQuery, StandardCharsets.UTF_8))
```
domivds: see previous comment

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/HttpClientUtils.java
```diff
@@ -0,0 +1,89 @@
+package zone.cogni.libs.sparqlservice.impl;
```
domivds: could go to a non-sparql util package

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/HttpClientUtils.java
```diff
@@ -0,0 +1,89 @@
...
+  private static final Logger log = LoggerFactory.getLogger(HttpClientUtils.class);
+
+  /**
+   * Creates a basic authenticator (username, password)
+   *
+   * @param username username
+   * @param password password (not null)
+   * @return authenticator
+   */
+  public static Authenticator createBasicAuthenticator(final String username,
```
domivds: is this basic as in "simple" or as in basic-auth

in first case, the name is confusing
in 2nd case, not sure this is related

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/StardogSparqlService.java
```diff
@@ -1,84 +1,77 @@
...
   @Override
   public void executeUpdateQuery(String updateQuery) {
-    Request request = Request
-            .Post(endpointUrl + "/update")
-            .setHeader(authHeader)
-            .body(new UrlEncodedFormEntity(Collections.singletonList(new BasicNameValuePair("update", updateQuery)), StandardCharsets.UTF_8));
-    executeAndConsume(request);
+    final HttpRequest request = HttpRequest
+        .newBuilder(URI.create(endpointUrl + "/update"))
+        .POST(HttpRequest.BodyPublishers.ofString("update=" + updateQuery, StandardCharsets.UTF_8))
```
domivds: see previous

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/GraphDBSparqlService.java
psiotwo: yes, utf-8

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/FusekiSparqlService.java
psiotwo: This code actually does not encode anything. Yet, the [Fuseki test was passing even before](https://github.com/cognizone/asquare/blob/beb6b9bdedbd1c30e932b3ab90ea3fd88b16fd74/cogni-libs/src/test/java/zone/cogni/libs/sparqlservice/impl/AbstractSparqlServiceTest.java#L98), without any encoding in place.

Anyway, I updated the code for percent-encoding the POST params for all service implementations.

psiotwo: This code actually does not encode anything. Yet, the [Fuseki test was passing even before](https://github.com/cognizone/asquare/blob/beb6b9bdedbd1c30e932b3ab90ea3fd88b16fd74/cogni-libs/src/test/java/zone/cogni/libs/sparqlservice/impl/AbstractSparqlServiceTest.java#L98), without any encoding in place.

Anyway, I updated the code for percent-encoding the POST params for all service implementations.

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
psiotwo: Fixed.

build.gradle
psiotwo: This is just a declaration of the version - `testImplementation` is at the actual dependency usage: https://github.com/cognizone/asquare/blob/beb6b9bdedbd1c30e932b3ab90ea3fd88b16fd74/build.gradle#L437

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
psiotwo: Ah true, thanks for pointing out. I missed that bit in the original code.

Will write some test for this too to avoid these kind of surprises ...

psiotwo: So JDK 11 HttpClient [does not seem to support digest auth](https://bugs.openjdk.org/browse/JDK-8285888). The authentication challenge needs to be implemented from scratch ...

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/HttpClientUtils.java
psiotwo: Looks like only Basic authentication is implemented in JDK client https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/376ab432aa0d120ace1077bc71bc694b1af86e0a/src/java.net.http/share/classes/jdk/internal/net/http/AuthenticationFilter.java#L276.

Anyway I removed the method as it is only used once and is not needed for now.

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
psiotwo: @domivds currently reimplemented to generic SPARQL 1.1. protocol calls using Apache HttpClient (only the Virtuoso service) bypassing Jena query execution - in case it turns out not to be performant enough, JDK11 HttpClient adjustments might be the way + added disabled integration tests for Virtuoso.

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
psiotwo: Not relevant anymore - using Apache HttpClient.

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
psiotwo: So Andy brought me to the right track: https://stackoverflow.com/questions/77901362/jena-4-connection-to-virtuoso 

Will try and update here.

psiotwo: OK, so rewired to `RDFConnection`, leaving out Apache HttpClient only in tests.

natan-cox-cognizone (commented):
ok, two things. somewhat related. not everything needs to be fixed in this PR. but lets discuss.

first, we completely switch using this implementation. not even a way of going back. maybe two services make sense? make a factorymethod could help? I dunno. this feels like pretty radical.

second, while switch the implementation might improve the performance (not even 100% if and how much) it will not fix the issue where a developer wants to use the same connection to do a number of operations until he is done. for this he needs a way of handling the transaction.

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
natan-cox-cognizone, line 1: ok, two things. somewhat related. not everything needs to be fixed in this PR. but lets discuss.

first, we completely switch using this implementation. not even a way of going back. maybe two services make sense? make a factorymethod could help? I dunno. this feels like pretty radical.

second, while switch the implementation might improve the performance (not even 100% if and how much) it will not fix the issue where a developer wants to use the same connection to do a number of operations until he is done. for this he needs a way of handling the transaction.

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
```diff
@@ -1,209 +1,119 @@
...
-      if (!HttpStatus.resolve(statusLine.getStatusCode()).is2xxSuccessful()) {
-        log.error("Virtuoso update failed with http status code {}", statusLine.getStatusCode());
-        String errorResponse = IOUtils.toString(response.getEntity().getContent(), StandardCharsets.UTF_8);
-        log.error("Virtuoso update failed with result {}", errorResponse);
-      }
-    }
-    catch (IOException ex) {
-      log.error("Virtuoso update failed {}", ex.getMessage());
-      throw new VirtuosoOperationException();
+    try (RDFConnection connection = getConnection()) {
```
natan-cox-cognizone: dont we need transaction management?

cogni-libs/src/main/java/zone/cogni/libs/core/utils/ApacheHttpClientUtils.java
```diff
@@ -0,0 +1,109 @@
...
+      final String password,
+      final boolean addBasicAuth,
+      final HttpEntity httpEntity,
+      boolean put,
+      final String contentType) {
+
+    try (final CloseableHttpClient httpclient = ApacheHttpClientUtils.buildHttpClient(username,
+        password)) {
+      final HttpEntityEnclosingRequestBase httpPost =
+          put ? new HttpPut(url) : new HttpPost(url);
```
natan-cox-cognizone: I don't want to be annoying but this starting on a newline because I reach a character limit is a little ugly, no?

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
natan-cox-cognizone: I have the feeling a wrapper could make sense (for the second case). I.e. it should expose RDFConnection and RdfStoreService and handle the transaction as a wrapper. 

Then it could make sense to keep the VirtuosoRdfStoreService "with connection" unaware of transactions btw. And also always "internal" i.e. not directly usable. 

Assume we have this
```
interface UpdateOperation {
    void execute(RdfStoreService rdfStoreService);
}
```

We could then implement this as wrapper

```
    public void update(UpdateOperation updateOp) { ... }
```
 
So, 
* wrapper creates a connection
* wrapper start transaction
* wrapper creates `ConnectionAwareVirtuosoRdfStoreService` (sorry... just kidding) as `rdfStoreService` variable
* wrapper invokes `updateOperation.execute(rdfStoreService)`
* wrapper handles catch/finally with rollback/commit

natan-cox-cognizone: So something like this ?

```
public class TransactedRdfStoreService implements RdfStoreService, AutoCloseable {

    private RDFConnection connection;
    private final RdfStoreService service;

    public TransactedRdfStoreService(...) {
...
    }

    public void update(UpdateOperation updateOp) {
        try {
            connection.begin(ReadWrite.WRITE);
            updateOp.execute(service);
            connection.commit();
        }  catch (Exception e) {
                connection.abort();
            log.error("Error executing update operation", e);
        } finally {
                connection.commit();
        }
    }

    @Override
    public void close() throws Exception {
            connection.close();
    }

    // Override other methods as they would delegate the call to the service instance...
}
```

and its use would be roughly 

```
try (TransactedRdfStoreService transactedService = new TransactedRdfStoreService(...)) {

    transactedService.update(rdfStoreService -> {
        // ... Perform your updates using the rdfStoreService ...
    });

}
```

cogni-libs/src/main/java/zone/cogni/libs/core/utils/ApacheHttpClientUtils.java
psiotwo: :-) OK, I try to reformat through intellij, maybe I forgot, will check. Anyway, is there any automated formatter+checkstyle for the project? Given the size, it would pay off IMO, as these issues might be frequent ...

psiotwo: @natan-cox-cognizone [here](https://github.com/cognizone/asquare/pull/229/commits/0548acadcef7411b67262f3dbd3ba2dc2a532799) you are.

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
psiotwo: > ok, two things. somewhat related. not everything needs to be fixed in this PR. but lets discuss.
> 
> first, we completely switch using this implementation. not even a way of going back. maybe two services make sense? make a factorymethod could help? I dunno. this feels like pretty radical.

@natan-cox-cognizone  [here](https://github.com/cognizone/asquare/pull/229/commits/0548acadcef7411b67262f3dbd3ba2dc2a532799) you are. I created just marker interface. Factory method would be possible, but not sure it pays off at this stage - it might make sense once there is some mechanism to set the default implementation through some env var / configuration. 

> second, while switch the implementation might improve the performance (not even 100% if and how much) it will not fix the issue where a developer wants to use the same connection to do a number of operations until he is done. for this he needs a way of handling the transaction.

True, good point. 

I can imagine creating a transactional decorator over `RdfStoreService`  (yr proposal above contains `public void update(UpdateOperation updateOp)` which I am not sure where it comes from ...). Yet, since the HttpClient implementation is back on the scene, I suggest not to bring it in this PR (it is already substantial).

access/src/main/java/zone/cogni/asquare/virtuoso/VirtuosoRdfStoreService.java
psiotwo: Transactionality should be ok, see https://jena.apache.org/documentation/rdfconnection/. Plus see the discussion above.

psiotwo: @natan-cox-cognizone [here](https://github.com/cognizone/asquare/pull/229/commits/0548acadcef7411b67262f3dbd3ba2dc2a532799) you are

cogni-libs/src/main/java/zone/cogni/libs/sparqlservice/impl/JenaModelSparqlService.java
natan-cox-cognizone: Ah. Well from the changes it did look like you purposely added the wrapper. If that is not the case then I would indeed keep it as it is. Seems, like there was a reason.

natan-cox-cognizone (approved):

domivds (approved):

PR #232, opened 2024-01-26
#T18654 escape new lines

christinaspachiou:
#T18654 escape new lines because updateLegalResourceImpactHistoricalImpact-en.sparql in casemates fails for migrated impactToLegalResourceComment or impactFromLegalResourceComment values which contain newLines

### DIFF
--- a/cogni-sem/src/main/java/zone/cogni/sem/jena/SparqlUtils.java
+++ b/cogni-sem/src/main/java/zone/cogni/sem/jena/SparqlUtils.java
@@ -1,13 +1,16 @@
 package zone.cogni.sem.jena;
 
-import org.apache.commons.lang3.StringUtils;
-
 public enum SparqlUtils {
   ;
 
-  public static String escapeLiteral(String literal) {
-    return StringUtils.replace(literal, "'", "\\'")
-                      .replace("\n", "\\n");
-
+  public static String escapeString(String literal) {
+    return literal.replace("\\", "\\\\") // Escape backslash, needs to be first otherwise it will escape the backslashes of the already escaped literals
+                  .replace("\t", "\\t")  // Escape tab
+                  .replace("\b", "\\b")  // Escape backspace
+                  .replace("\n", "\\n")  // Escape newline
+                  .replace("\r", "\\r")  // Escape carriage return
+                  .replace("\f", "\\f")  // Escape formfeed
+                  .replace("'", "\\'")   // Escape single quote
+                  .replace("\"", "\\\""); // Escape double quote
   }
 }
--- a/cogni-sem/src/main/java/zone/cogni/sem/jena/SparqlUtils.java
+++ b/cogni-sem/src/main/java/zone/cogni/sem/jena/SparqlUtils.java
@@ -6,7 +6,8 @@ public enum SparqlUtils {
   ;
 
   public static String escapeLiteral(String literal) {
-    return StringUtils.replace(literal, "'", "\\'");
+    return StringUtils.replace(literal, "'", "\\'")
+                      .replace("\n", "\\n");
 
   }
 }
--- a/cogni-sem/src/test/java/zone/cogni/sem/jena/SparqlUtilsTest.java
+++ b/cogni-sem/src/test/java/zone/cogni/sem/jena/SparqlUtilsTest.java
@@ -5,39 +5,54 @@ import org.junit.jupiter.api.Test;
 import zone.cogni.libs.sparqlservice.impl.JenaModelSparqlService;
 
 public class SparqlUtilsTest {
+  private static final String URI = "https://fedlex.data.admin.ch/eli/oc/2023/423/legal-analysis";
+  private static final String GRAPH_URI = URI + "/graph";
   private final JenaModelSparqlService jenaModelSparqlService = new JenaModelSparqlService();
 
   @Test
   public void test() {
-    initializeSource();
-    confirmResult();
+    testEscape("A\ncomment"); // test newline
+    testEscape("A\\comment"); // test backslash
+    testEscape("A\bcomment"); // test backspace
+    testEscape("I'mAComment"); // test single quote
+    testEscape("A\fcomment"); // test form feed
+    testEscape("A\tcomment");  // test tab
+    testEscape("A\rcomment"); // test carriage return
+    testEscape("com\"ment"); // test double quotes
   }
 
-  private void initializeSource() {
+  private void testEscape(String value) {
+    initializeSource(value);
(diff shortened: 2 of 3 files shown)


cogni-sem/src/main/java/zone/cogni/sem/jena/SparqlUtils.java
```diff
@@ -6,7 +6,8 @@ public enum SparqlUtils {
   ;
 
   public static String escapeLiteral(String literal) {
-    return StringUtils.replace(literal, "'", "\\'");
+    return StringUtils.replace(literal, "'", "\\'")
+                      .replace("\n", "\\n");
```
natan-cox-cognizone: hmm. yeah. there are chars that need to be escaped in general. if it is a normal string and you want to do the normal Java escaping because you need an escaped string. this is probably not the way.

natan-cox-cognizone: `public static String escapeString(String input) {
    return input
            .replace("\\", "\\\\") // Escape backslash
            .replace("\t", "\\t")  // Escape tab
            .replace("\b", "\\b")  // Escape backspace
            .replace("\n", "\\n")  // Escape newline
            .replace("\r", "\\r")  // Escape carriage return
            .replace("\f", "\\f")  // Escape formfeed
            .replace("\'", "\\\'") // Escape single quote
            .replace("\"", "\\\""); // Escape double quote
}
`

is it this you want?
if so, 
`import org.apache.commons.text.StringEscapeUtils;

public static String escapeString(String input) {
    return StringEscapeUtils.escapeJava(input);
}`

or even remove this function and just use the method above.

domivds: @natan-cox-cognizone it's the method to escape literals when constructing sparql's
i think the escapeJava stuff will be a bit too much and might alter the final value (or not), so we need to come up with the list of values we want to escape

@christinaspachiou make a unittest for this with a dummy dataset that uses this with insert queries and then just fetch the data via the dataset (model) to see if what we have in the store is what we want

natan-cox-cognizone: ok, thanks for the clarification. 
then we should go to the SPARQL spec or so, no? We are allowed to do both "literal" and 'literal'. So, I am guessing depending on the wrapping using "" or '' we should also escape " or '

natan-cox-cognizone: https://www.w3.org/TR/sparql11-query/#grammarEscapes ?

cogni-sem/src/test/java/zone/cogni/sem/jena/SparqlUtilsTest.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  private void initializeSource() {
+    jenaModelSparqlService.executeUpdateQuery(
+      "PREFIX jolux: <http://data.legilux.public.lu/resource/ontology/jolux#> " +
+      "INSERT DATA { " +
+      "  GRAPH <https://fedlex.data.admin.ch/eli/oc/2023/423/legal-analysis/graph> { " +
+      "    <https://fedlex.data.admin.ch/eli/oc/2023/423/legal-analysis> a jolux:LegalAnalysis. " +
+      "    <https://fedlex.data.admin.ch/eli/oc/2023/423/legal-analysis> jolux:impactFromLegalResourceComment " + "'" + SparqlUtils.escapeLiteral("A\ncomment") + "'" +
+      "  }" +
```
domivds: check the doc natan sent, and do more tests based on that
also, just any uri with just 1 literal is needed
and i would than have a method that receives the string, clears the dataset, adds it, re-fetches it and asserts

your test method can run a bunch of tests for the different to-escape characters

cogni-sem/src/main/java/zone/cogni/sem/jena/SparqlUtils.java
domivds: @christinaspachiou check above link for what to escape

cogni-sem/src/test/java/zone/cogni/sem/jena/SparqlUtilsTest.java
```diff
@@ -0,0 +1,58 @@
...
+    confirmResult(value);
+  }
+
+  private void initializeSource(String value) {
+    jenaModelSparqlService.executeUpdateQuery(
+      "PREFIX jolux: <http://data.legilux.public.lu/resource/ontology/jolux#> " +
+      "INSERT DATA { " +
+      "  GRAPH <" + GRAPH_URI + "> { " +
+      "    <" + URI + "> a jolux:LegalAnalysis. " +
+      "    <" + URI + "> jolux:impactFromLegalResourceComment \"" + SparqlUtils.escapeString(value) + "\"" +
```
domivds: @christinaspachiou did you validate the fact that you define a literal with single quote also?
would be good to do the tests for both cases (single quote and double quote)

cogni-sem/src/test/java/zone/cogni/sem/jena/SparqlUtilsTest.java
```diff
@@ -0,0 +1,58 @@
...
+import zone.cogni.libs.sparqlservice.impl.JenaModelSparqlService;
+
+public class SparqlUtilsTest {
+  private static final String URI = "https://fedlex.data.admin.ch/eli/oc/2023/423/legal-analysis";
+  private static final String GRAPH_URI = URI + "/graph";
+  private final JenaModelSparqlService jenaModelSparqlService = new JenaModelSparqlService();
+
+  @Test
+  public void test() {
+    testEscape("A\ncomment"); // test newline
```
domivds: add "mixed" testcases 
things like this: "Bla\\n" and "Bla\\\n"

cogni-sem/src/test/java/zone/cogni/sem/jena/SparqlUtilsTest.java
christinaspachiou: yes I had tested, but I added that here too

cogni-sem/src/test/java/zone/cogni/sem/jena/SparqlUtilsTest.java
christinaspachiou: added a couple of them

domivds (approved):

natan-cox-cognizone (approved):

PR #234, opened 2024-04-10
T19191/content lib update

christinaspachiou:
@natan-cox-cognizone we will introduce a join field to create a parent child functionality in our elastic documents. This will be used for content searching in legiswiss. it is need it to add a top level object in the elastic docs to define the relation in the parent document. I kind of tried to make it work similarly as the facets.


async-indexing/src/main/java/zone/cogni/asquare/service/index/GraphIndexService.java
```diff
@@ -84,6 +86,11 @@ public Boolean indexUriSync(ResourceIndex resourceIndex, ApplicationView view, P
         }
       }
 
+      if (parentChildSupplier != null) {
+        ObjectNode joinFieldNode = parentChildSupplier.apply(resourceIndex.getType());
+        json.set("my_join_field", joinFieldNode);
+      }
```
domivds: the idea was to have a generic interceptor
You can just make a `Consumer<ObjectNode>`, name can be something like `postIndexInterceptor`

service/src/main/java/zone/cogni/asquare/service/index/IndexConfigProvider.java
```diff
@@ -23,35 +23,39 @@ public class IndexConfigProvider {
...
   private final ElasticStore elasticStore;
   private final Function<ResourceIndex, Function<TypedResource, ObjectNode>> facetConversionSupplier;
+  private final Function<String, ObjectNode> parentChildSupplier;
 
   public IndexConfigProvider(RdfStoreService rdfStoreService,
                              Function<ResourceIndex, ApplicationProfile> applicationProfileSupplier,
                              Elasticsearch7Store elasticStore,
-                             Function<ResourceIndex, Function<TypedResource, ObjectNode>> facetConversionSupplier) {
+                             Function<ResourceIndex, Function<TypedResource, ObjectNode>> facetConversionSupplier,
+                             Function<String, ObjectNode> parentChildSupplier) {
```
domivds: seems you changed all constructors, this will break all current projects using this!

As we have lombok you can easily add a builder, annotate the optional fields with `@Nullable`. That way we will not have to pass all null's to the contructor

async-indexing/src/main/java/zone/cogni/asquare/service/index/GraphIndexService.java
domivds: Also, put the 2 if blocks each in a separate method

domivds (changes requested):

async-indexing/src/main/java/zone/cogni/asquare/service/index/GraphIndexService.java
christinaspachiou: done

service/src/main/java/zone/cogni/asquare/service/index/IndexConfigProvider.java
christinaspachiou: ok, its reverted and I use the builder in casemates

async-indexing/src/main/java/zone/cogni/asquare/service/index/GraphIndexService.java
```diff
@@ -109,6 +104,24 @@ else if(StringUtils.isNotBlank(resourceIndex.getGraph())) {
     return false;
   }
 
+  private void executeFacetConversion(ResourceIndex resourceIndex, TypedResource resource, ObjectNode json) {
+    if (facetConversionSupplier != null) {
```
natan-cox-cognizone: I do not like this. See for example the book "Refactoring" -> https://silab.fon.bg.ac.rs/wp-content/uploads/2016/10/Refactoring-Improving-the-Design-of-Existing-Code-Addison-Wesley-Professional-1999.pdf

Lookup chapter: Replace Nested Conditional with Guard Clauses
A method has conditional behavior that does not make clear the normal path of execution.

async-indexing/src/main/java/zone/cogni/asquare/service/index/GraphIndexService.java
```diff
@@ -109,6 +104,24 @@ else if(StringUtils.isNotBlank(resourceIndex.getGraph())) {
...
+        ObjectNode facets = facetConversion.apply(resource);
+        if (facets != null) {
+          json.set("facets", facets);
+        }
+      }
+    }
+  }
+
+  private void executePostIndexInterceptor(ObjectNode json, String type) {
+    if (postIndexInterceptor != null) {
```
natan-cox-cognizone: same here

async-indexing/src/main/java/zone/cogni/asquare/service/index/GraphIndexService.java
christinaspachiou: i changed it to make it more readable

natan-cox-cognizone (approved):

service/src/main/java/zone/cogni/asquare/service/index/IndexConfigProvider.java
```diff
@@ -13,45 +15,54 @@
...
-                             Elasticsearch7Store elasticStore,
-                             Function<ResourceIndex, Function<TypedResource, ObjectNode>> facetConversionSupplier) {
+  @Nullable
+  private final BiConsumer<ObjectNode, String> postIndexInterceptor;
+
+  public IndexConfigProvider( RdfStoreService rdfStoreService,
+                              Function<ResourceIndex, ApplicationProfile> applicationProfileSupplier,
+                              Elasticsearch7Store elasticStore,
+                             @Nullable Function<ResourceIndex, Function<TypedResource, ObjectNode>> facetConversionSupplier,
+                             @Nullable BiConsumer<ObjectNode, String> postIndexInterceptor) {
```
domivds: isn't this the `@AllArgsConstructor` ?
and the other constructor is gone, so we still have the backwards compatible issue....

note, as we have a builder now, you can set the other constructors to deprecated and make the `@AllArgsConstructor` private

service/src/main/java/zone/cogni/asquare/service/index/IndexConfigProvider.java
```diff
@@ -70,6 +81,10 @@ public Function<ResourceIndex, Function<TypedResource, ObjectNode>> getFacetConv
     return facetConversionSupplier;
   }
 
+  public BiConsumer<ObjectNode, String> getPostIndexInterceptor() {
+    return postIndexInterceptor;
+  }
```
domivds: Can't we just add `@Getter` ?

domivds (changes requested):

domivds (approved):

PR #238, opened 2025-07-09
#T20838 upgrade gradle/libs + code changes

christinaspachiou:


CHANGELOG.md
```diff
@@ -1,2 +1,14 @@
...
+## [0.8.0] - 2025-07-09
+
+### Changed
+- Upgraded from Springboot 2.7.18 to Springboot 3.5.3
+- Upgraded from Java 11 to Java 17
+- Upgraded gradle to 7.6.4 to be compatible with Java 17
+- Upgraded from Jena 4 to Jena 5
+- Upgraded rest of libs to match the new java/spring versions
+
+- Moved the index package from service to async-indexing
```
natan-cox-cognizone: this might not be so useful. i realized the dependencies are different than I assumed. sorry.

build.gradle
```diff
@@ -3,25 +3,34 @@ import pl.allegro.tech.build.axion.release.domain.scm.ScmPosition
...
-    guavaVersion = '30.0-jre'
-    httpComponentsVersion = '4.5.13'
-    jenaVersion = '4.10.0'
-    springBootVersion = '2.7.18'
-    springSecurityVersion = '5.8.9'
-    springSecuritySaml2CoreVersion = '1.0.10.RELEASE'
+    guavaVersion = '33.4.8-jre'
+    httpclient5Version = "5.5"
+    httpComponentsVersion = '4.5.14'
+    jacksonVersion = "2.19.1"
```
natan-cox-cognizone: why the full list of deps here? if they are implicitly managed by spring boot this is better. because now we have to manually update all versions, no? @domivds

natan-cox-cognizone: ah... most of it is moving the version from below to here. ignore!

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-all.zip
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.4-all.zip
```
natan-cox-cognizone: no higher?

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/jenamemory/LocalTdbRdfStoreService.java
```diff
@@ -83,7 +81,7 @@ public LocalTdbRdfStoreService(
       this.tdbLocation, this.firstResultTimeout, this.firstResultTimeUnit, this.overallTimeout, this.overallTimeUnit
     );
 
-    this.dataset = TDBFactory.createDataset(tdbLocationFolder.getAbsolutePath());
+    this.dataset = TDB2Factory.connectDataset(tdbLocationFolder.getAbsolutePath());
```
natan-cox-cognizone: hmm. moving from TDB to TDB2 might have significant impact. are these formats binary compatible? if not then someone upgrading might just destroy their TDB database!

natan-cox-cognizone (commented):
upgrade to Jena 5 looks pretty significant. we should definitely test this thorougly!!

alexey-lukashov (commented):
@christinaspachiou this is big update javax->jakarta, java 11->17, jena 4->5, spring boot 2->3, I believe it compiles but the important question - was this version tested on any project?

cogni-sem/src/main/java/zone/cogni/sem/jena/JenaUtils.java
```diff
@@ -255,12 +257,11 @@ private static RDFReaderI getBasicReader(Model model, Resource resource) {
...
 
   private static RDFReaderI getReaderByRdfSyntax(Model model, String language) {
-    try {
-      return model.getReader(language);
-    }
-    catch (IllegalStateException ignored) {
-      return model.getReader();
+    Lang lang = RDFLanguages.nameToLang(language);
+    if (lang == null) {
+      return model.getReader("RDF/XML"); // default lang that was used in jena 4
```
psiotwo: There is the Lang enum

build.gradle
```diff
@@ -3,25 +3,34 @@ import pl.allegro.tech.build.axion.release.domain.scm.ScmPosition
 
 buildscript {
   ext {
-    apachePoiVersion = '5.2.2'
-    commonsLang3Version = '3.11'
-    commonsCodecVersion = '1.15'
-    commonsIoVersion = '2.11.0'
+    apachePoiVersion = '5.4.1'
```
psiotwo: TOML file is a convenient way of managing versions.

psiotwo (commented):
Minor comments.

But should be tested - not sure if there is an app to test it in detail. ... Automated tests would be the best OFC.

alexey-lukashov (dismissed):

gradle/wrapper/gradle-wrapper.properties
christinaspachiou: Indeed it's possible

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/jenamemory/LocalTdbRdfStoreService.java
christinaspachiou: @natan-cox-cognizone indeed it seems they are not compatible reading this: https://jena.apache.org/documentation/tdb2/tdb2_migration.html
If im not wrong we need to reload a backup of the data in the tdb2 location? 
I only changed it to tdb2 because tdb1 only works with jena 4

cogni-sem/src/main/java/zone/cogni/sem/jena/JenaUtils.java
christinaspachiou: done

build.gradle
christinaspachiou: introduced toml

christinaspachiou:
> Minor comments.
> 
> But should be tested - not sure if there is an app to test it in detail. ... Automated tests would be the best OFC.

@psiotwo and @natan-cox-cognizone I didnt dive too deep in the jena upgrades as it's not a library used by legilux. In the near future we will be upgrading lux-casemates which uses a lot these parts of asquare so thorough testing needs be done there. I can write a few unit tests if you want but ideally we agree on which modules we want to have tested

alexey-lukashov (commented):
TDB -> TDB2 seems risky

triplestore-jena-memory/src/main/java/zone/cogni/asquare/triplestore/jenamemory/LocalTdbRdfStoreService.java
natan-cox-cognizone: ok, just to make migration a bit more explicit I would remove LocalTdbRdfStoreService and introduce LocalTdb2RdfStoreService. (note: I guess a rename is fine)

That way people upgrading to Jena 5 will get a compilation error. Also add a big comment in the LocalTdb2RdfStoreService so people get a warning. 
Maybe even a comment that this is not fully tested yet (since you do not use it in Legilux).

Finally, we will tell the teams that use TDB that Jena 5 cannot work with TDB and it needs a DB migration to TDB2.

christinaspachiou: @natan-cox-cognizone as discussed I removed things related to LocalTdbRdfStore and also moved back the index package to service module

natan-cox-cognizone (approved):
from my point of view it is fine.

PR #245, opened 2025-10-23
Feature/local tdb store

marianna-tselepi:
- put back classes removed on https://github.com/cognizone/asquare/pull/238
- update code for jena5 & TDB1 compatibility

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -616,6 +616,7 @@ project('triplestore-jena-memory') {
     api project(':triplestore')
 
     // libs
+    api libs.jena.tdb1
     api libs.jena.tdb2
 
     // TESTING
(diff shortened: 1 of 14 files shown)


christinaspachiou (approved):

alexey-lukashov (approved):

psiotwo (approved):
weak lgtm

I assume most of the stuff is just revert of the previous classes (that worked), right? Otherwise I would have some comments there ...

christinaspachiou:
@marianna-tselepi maybe we should update also the the https://github.com/cognizone/asquare/blob/develop/CHANGELOG.md ?

marianna-tselepi:
> I assume most of the stuff is just revert of the previous classes (that worked), right? Otherwise I would have some comments there ...

@psiotwo most of the stuff was reverted yes, I have only updated the code to make it compatible with jena 5 and added some extra tests. any comment (or github issue) is welcome.
I'll merge this PR since it's blocking point, and address all the improvements (if any) in a separate PR. @christinaspachiou same applies for the changelog

PR #252, opened 2026-03-12
upgrade to java 21 and to spring boot 4.0.2

gementakis:


gementakis:
This PR was supposed Jena5 and https://github.com/cognizone/asquare/pull/254 was for Jena6. We now close this and keep only the other one.

PR #255, opened 2026-04-15
Feature/T27908 update asquare to java 21 and spring boot 3 and jena 6

gementakis:


cogni-sem/src/main/java/zone/cogni/sem/jena/template/JenaQueryTemplate.java
```diff
@@ -195,6 +197,10 @@ public static QueryExecution newQueryExecution(Model model, Query sparql) {
   }
 
   public static QueryExecution newQueryExecution(Model model, Query sparql, QuerySolution querySolution) {
```
domivds: nice to have: as `querySolution` can be `null` add `@Nullable` to method signature

cogni-sem/src/main/java/zone/cogni/sem/jena/template/JenaQueryUtils.java
```diff
@@ -67,7 +67,11 @@ public static QueryExecution newQueryExecution(Model model, String sparql, Query
   }
 
   public static QueryExecution newQueryExecution(Model model, Query sparql, QuerySolution querySolution) {
```
domivds: nice to have: as querySolution can be null add @Nullable to method signature

build.gradle
```diff
@@ -182,7 +176,7 @@ subprojects {
 
 idea {
   project {
-    languageLevel = "1.8"
+    languageLevel = "17"
```
domivds: shouldn't this also be 21?

gementakis:
Feature will be delivered before java/asquare/jena upgrades.

